PR #232, opened 2026-07-05
docs(skill): define Kungfu Skill architecture

dongkeren:
## Summary
- add ADR-0015 for Kungfu Skill as the agent context layer above kfx
- add docs/skills.md covering SKILL.md minimal source, catalog/envelope injection, Node/Python manage modes, and kfx dependency composition
- link the new design from README, docs/MAP.md, and ADR index

## Validation
- git diff --check

## Risk
- documentation only; implementation is explicitly marked as not yet landed


kungfu-origin (approved):
Approved: documentation-only architecture decision and design target for Kungfu Skill. Validation noted: git diff --check.

PR #239, opened 2026-07-05
docs: state fact-first product goal

dongkeren:
## Summary
- state the product goal in README: make fact-first responsibility the path of least resistance
- add a Product Goal section to docs/facts-before-trust.md
- connect the goal from docs/design-philosophy.md

## Validation
- git diff --check
- checked the public docs patch for Atlas or AI-maintenance leakage terms


kungfu-origin (approved):
Reviewed docs-only product goal anchor. Public wording stays product-facing and keeps private Atlas context out.

PR #240, opened 2026-07-05
fix(deps): update vulnerable toolchain packages

dongkeren:
## Summary
- update Python dependency floors for current advisory fixes in framework/core
- upgrade GUI build tooling to Vite 6 and electron-vite 5
- move vulnerable tar/js-yaml transitive edges to patched versions through pnpm workspace overrides
- refresh uv and pnpm lockfiles

## Validation
- `uv lock --check`
- `./kungfu-code audit --audit-level low` reports no known vulnerabilities
- `git diff --check`
- package JSON parse check for root, framework/core, and framework/gui manifests

## Known blocker
- `./kungfu-code install --frozen-lockfile --lockfile-only` still fails because `framework/core/package.json` references unpublished optional packages: `@kungfu-tech/core-darwin-arm64`, `@kungfu-tech/core-linux-x64`, and `@kungfu-tech/core-win32-x64` at `4.0.0-alpha.0`. This is separate from the vulnerability updates and blocks full `build:app` validation in the current local mirror/registry state.

## Risk
- Vite/electron-vite and Python packaging tools move across major versions; follow-up GUI/core build validation is needed after the optional platform package issue is resolved.


kungfu-origin (approved):
Dependency update looks scoped and validation checks are passing.

PR #298, opened 2026-07-06
perf(core): drop scientific stack from the frozen runtime

dongkeren:
## What

Stop force-including the scientific stack in the frozen `kungfu` runtime.

`kungfu_cli.py` force-included `numpy`/`pandas`/`plotly`, and Nuitka
additionally pulled in `scipy`/`statsmodels`/`botocore`/`boto3` as
transitive weight. None of these are used on the core runtime path:

- `plotly`, `scipy`, `statsmodels`, `botocore`, `boto3` — zero imports across the tree (legacy quant declarations).
- `numpy`/`pandas` — only used by `journal sessions` to render a table.

`find_sessions` is rewritten to return plain dicts and the table is
rendered with the existing `tabulate` dependency, so the six packages
can move from `--include-package` to `--nofollow-import-to`.

## Size (macOS arm64, freeze + package:app, C++ core unchanged)

| | before | after | saved |
|---|--:|--:|--:|
| frozen main binary | 257M | 22M | -235M |
| `dist/kungfu` | 709M | 378M | -331M |
| `Kungfu.app` | 993M | 699M | **-294M (-30%)** |

## Validation

Freeze + `package:app` rebuilt on top of an unchanged native core.
Runtime smoke tests green: `journal sessions` (now stdlib-only),
`journal sessions -s duration -a`, `kfx contract`, `config show`,
`skill`, `--version`.

## Follow-up

On-demand heavy dependencies should ship with the kfx that needs them
(`developer/sdk` python-AOT), keeping the core a lean python runtime.

### DIFF
--- a/framework/core/src/python/kungfu/cli/commands/journal.py
+++ b/framework/core/src/python/kungfu/cli/commands/journal.py
@@ -88,22 +88,25 @@ def journal(ctx, mode, category, group, name):
 )
 @journal_command_context
 def sessions(ctx, sortby, ascending, tablefmt):
-    all_sessions = kfj.find_sessions(ctx).sort_values(by=sortby, ascending=ascending)
-    all_sessions["begin_time"] = all_sessions["begin_time"].apply(
-        lambda t: kft.strftime(t, kft.SESSION_DATETIME_FORMAT)
+    all_sessions = sorted(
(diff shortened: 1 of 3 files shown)


dongkeren:
Closing as superseded by dev/v4/v4.0: the scientific-stack frozen-runtime work is already present via 74ac370d8 and a76e7d9ab.

PR #315, opened 2026-07-07
ci: grant buildchain build issue permission

dongkeren:
## Summary
- grant the Buildchain reusable build workflow the issue permission it requests
- unblock alpha release-candidate build startup while keeping publish/release options disabled in Buildchain config

## Validation
- actionlint .github/workflows/build.yml .github/workflows/release-verify.yml .github/workflows/release-new-version.yml .github/workflows/buildchain-validate.yml
- git diff --check

## Build policy
- This PR targets dev/v4/v4.0, so it should only run signoff/validate and must not run the product Build workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ on:
 permissions:
   contents: read
   pull-requests: read
+  issues: write
   id-token: write
 
 jobs:


kungfu-origin (approved):
Approved: narrow permission fix for the Buildchain reusable workflow caller.

PR #328, opened 2026-07-07
ci: prepare buildchain source build prerequisites

dongkeren:
## Summary
- keep Buildchain no-optional installs while injecting only the current libnode platform package for source rebuilds
- mark Buildchain source builds and auto-detect a Conan default profile when missing
- add common Windows user tool directories to PATH for lifecycle commands

## Validation
- node --check scripts/buildchain-run-kungfu-code.mjs artifact/scripts/dist.mjs framework/core/.gyp/run-conan.js
- pnpm exec biome check scripts/buildchain-run-kungfu-code.mjs artifact/scripts/dist.mjs framework/core/.gyp/run-conan.js
- pnpm run check:types
- node scripts/buildchain-run-kungfu-code.mjs install --frozen-lockfile --no-optional
- libnode NODE_PATH probe with @kungfu-tech/libnode-darwin-arm64@22.22.3-kf.1

### DIFF
--- a/artifact/scripts/dist.mjs
+++ b/artifact/scripts/dist.mjs
@@ -7,6 +7,7 @@
 
 import { spawnSync } from 'node:child_process';
 import fs from 'node:fs';
+import { createRequire } from 'node:module';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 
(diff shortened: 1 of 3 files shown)


kungfu-origin (approved):
Buildchain source-build prerequisite fix looks scoped and validated.

PR #330, opened 2026-07-07
ci: align core optional lockfile specifiers

dongkeren:
## Summary
- add the framework/core optional platform-package specifiers to the lockfile importer so Buildchain no-optional frozen installs validate on clean runners

## Validation
- CI=true pnpm install --frozen-lockfile --no-optional --lockfile-only
- node scripts/buildchain-run-kungfu-code.mjs install --frozen-lockfile --no-optional
- git diff --check


kungfu-origin (approved):
Lockfile importer alignment is minimal and validated.

PR #332, opened 2026-07-07
ci: discover uv from windows runner users

dongkeren:
## Summary
- add Windows user tool directories from C:\Users\* to the Buildchain wrapper PATH so self-hosted runner jobs can find an existing uv.exe during source rebuilds
- keep this scoped to Windows Buildchain lifecycle execution; no tool installation is performed

## Validation
- node --check scripts/buildchain-run-kungfu-code.mjs
- pnpm exec biome check scripts/buildchain-run-kungfu-code.mjs
- node scripts/buildchain-run-kungfu-code.mjs exec node -e "console.log(process.env.KUNGFU_BUILDCHAIN_NO_OPTIONAL, process.env.KUNGFU_BUILDCHAIN_SOURCE_BUILD)"
- git diff --check

### DIFF
--- a/scripts/buildchain-run-kungfu-code.mjs
+++ b/scripts/buildchain-run-kungfu-code.mjs
@@ -21,6 +21,24 @@ function existingDirs(dirs) {
   return dirs.filter((dir) => dir && fs.existsSync(dir));
 }
 
+function windowsUserToolPathDirs() {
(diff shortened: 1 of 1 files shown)


kungfu-origin (approved):
Windows runner uv path discovery is scoped and validated.

PR #336, opened 2026-07-07
alpha: build v4.0.0-alpha.0

dongkeren:
Fresh alpha promotion attempt after dev fixes through #335.

One-PR-one-Build guard:
- This PR is the single Build attempt for dev head 68ae2cd19d4c376fef5d53e6a41ba68735389bbf.
- If this Build fails, close this PR and fix forward through dev before opening another alpha PR.

Previous alpha PR #333 was closed after its single Build run failed and the remaining queued jobs were cancelled.

PR #337, opened 2026-07-07
ci: bootstrap uv for buildchain windows

dongkeren:
## Summary
- upgrade @kungfu-tech/buildchain to the current latest release package, 2.8.8
- bootstrap uv into .buildchain/uv on Windows only when uv is not already available
- keep the bootstrap scoped to the lifecycle wrapper instead of mutating runner-global state
- exclude generated .buildchain runtime files from the no-bash guard used by verify

## Evidence from alpha #336
- Windows now installs libnode and Rollup platform packages but still fails because uv is unavailable on the runner
- Linux builds the AppImage and reports Buildchain observability events, then fails verify because .buildchain/runtime/scripts/check-workflows.sh is generated by Buildchain runtime and should not be scanned as repo source

## Validation
- node --check scripts/buildchain-run-kungfu-code.mjs
- node --check scripts/no-bash-guard.mjs
- node scripts/no-bash-guard.mjs
- ./kungfu-code exec biome check scripts/buildchain-run-kungfu-code.mjs scripts/no-bash-guard.mjs artifact/package.json developer/sdk/package.json
- ./kungfu-code run check:types
- CI=true ./kungfu-code install --frozen-lockfile --no-optional --lockfile-only

### DIFF
--- a/artifact/package.json
+++ b/artifact/package.json
@@ -48,6 +48,6 @@
     "url": "https://github.com/kungfu-systems/kungfu.git"
   },
   "devDependencies": {
-    "@kungfu-tech/buildchain": "2.8.7"
+    "@kungfu-tech/buildchain": "2.8.8"
   }
 }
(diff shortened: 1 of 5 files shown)


kungfu-origin (approved):
Approved: scoped CI/runtime fix based on #336 failure logs; local validation passed.

PR #344, opened 2026-07-07
feat(gui): add integrated window chrome

dongkeren:
## Summary
- add a shared renderer titlebar for Electron GUI shell chrome
- use macOS hidden inset titlebar with traffic-light safe area
- use Windows custom frame controls and keep Linux on conservative native chrome
- add a command/search entry that opens enabled KFX views by id or title

## Validation
- ./kungfu-code --filter @kungfu-tech/gui run build
- ./kungfu-code check
- ./kungfu-code build
- ./kungfu-code product gui dev (macOS visual smoke confirmed)

## Notes
- Windows and Linux behavior is implemented by platform branches but only build-validated from macOS in this pass.

### DIFF
--- a/framework/gui/src/main/index.ts
+++ b/framework/gui/src/main/index.ts
@@ -22,6 +22,9 @@ import {
   HIDE_CHANNEL,
   SET_BOUNDS_CHANNEL,
   SHOW_CHANNEL,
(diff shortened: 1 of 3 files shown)


kungfu-origin (approved):
Reviewed integrated window chrome implementation and local validation output.

PR #361, opened 2026-07-07
alpha: build v4.0.0-alpha.0

dongkeren:
Source dev/v4/v4.0 @ 5002e08b53126bdd371e66e052a822b241a44def.

Purpose: retry Buildchain alpha product build after fixing Windows pinned Node lifecycle and tolerating the Windows uv mount-point interpreter discovery error in the mypy probe only for that specific error.

Guard: one alpha PR, one Build run; if the Build run fails, close this PR and cancel remaining jobs before fixing forward through dev.


dongkeren:
Closing after the single allowed Build run failed on Windows. Cancelling remaining jobs before fixing forward through dev.

PR #372, opened 2026-07-07
ci: upload only staged release artifacts

dongkeren:
## Summary
- stage top-level electron-builder outputs from artifact/dist into artifact/release
- upload artifact/release in the Buildchain reusable build instead of the full artifact/dist tree
- ignore generated artifact/release locally

## Verification
- node --check artifact/scripts/dist.mjs
- ./kungfu-code check

## Risk
- The Buildchain action scans concrete paths, not glob expressions, so this keeps a directory target while excluding unpacked app directories generated under artifact/dist.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
       runner-preset: kungfu-v4-self-hosted
       artifact-name: kungfu
       artifact-paths: |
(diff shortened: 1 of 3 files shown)


kungfu-origin (approved):
Approved: CI artifact upload is narrowed to staged release files while preserving generated dist output.

PR #395, opened 2026-07-08
refactor(core): fence node longfist public registry

dongkeren:
## Summary
- bind Node Longfist public `types` and `carrierTypes` to `CorePublicDataTypes`
- update the status view wording to show the core registry rather than the legacy compiled registry
- extend the yijinjing greenfield guard to block public Node binding and CorePublic* registry regressions

## Validation
- ./kungfu-code build
- ./kungfu-code check
- node scripts/check-yijinjing-greenfield.mjs --all
- node scripts/check-carrier-action-envelope.mjs --all
- git diff --check
- runtime probe: Longfist public types/carrierTypes expose 27 core entries and no legacy trading/profile entries


kungfu-origin (approved):
Public Node longfist registry is now fenced to the v4 core surface, and the greenfield guard covers the regression.

PR #396, opened 2026-07-08
refactor(core): name legacy compiled longfist registry

dongkeren:
## Summary
- add explicit LegacyCompiled* registry names for internal compiled longfist compatibility decode paths
- move journal replay, console, typed dump, rx custom-event detection, and Node Frame typed decode away from direct AllTypes/AllDataTypes names
- extend greenfield/dependency guards so public bindings and future yijinjing code cannot reintroduce legacy registry coupling

## Validation
- ./kungfu-code build
- ./kungfu-code check
- node scripts/check-carrier-action-envelope.mjs --all
- node scripts/check-yijinjing-greenfield.mjs --all
- git diff --check


kungfu-origin (approved):
Reviewed staged registry naming and greenfield guards; validation evidence is sufficient.

PR #397, opened 2026-07-08
refactor(core): remove watcher legacy data path

dongkeren:
## Summary
- remove Watcher legacy trading-data reader/cache/sync path and old command shim exports
- slim Watcher constructor/config/API to neutral runtime arguments
- drop LegacyRefreshDataTypes/Tags and update ADR plus bench usage

## Validation
- ./kungfu-code check
- ./kungfu-code build
- git diff --check
- node --check framework/core/lib/kungfu.js
- node --check framework/core/tests/bench/dispatch_watcher_bench.js
- rg residual scan for removed legacy watcher symbols/config names


kungfu-origin (approved):
Reviewed watcher legacy path removal and validation evidence.

PR #400, opened 2026-07-08
refactor(core): rename libkungfu runtime sources

dongkeren:
## Summary
- move libkungfu implementation sources from src/yijinjing to src/runtime
- update the libkungfu CMake cache glob to the new runtime path
- update docs links that referenced the old physical source directory

## Notes
- public include paths remain <kungfu/yijinjing/...>
- C++ namespaces remain kungfu::yijinjing
- Python/Node public APIs are unchanged

## Validation
- ./kungfu-code check
- ./kungfu-code build
- git diff --check
- rg old src/yijinjing path references


kungfu-origin (approved):
Reviewed physical runtime source directory rename and validation evidence.

PR #405, opened 2026-07-08
refactor(core): split yijinjing platform utilities

dongkeren:
## Summary
- move runtime-only stacktrace, terminal, signal, and Windows AppContainer APIs from libyijinjing into libkungfu runtime
- keep yijinjing focused on hash and mmap primitives plus journal/storage contracts
- tighten the yijinjing dependency guard and embedding docs so old util includes cannot return

## Validation
- ./kungfu-code fix
- ./kungfu-code check
- ./kungfu-code build
- node framework/core/slices/embedding/run.mjs
- git diff --check


kungfu-origin (approved):
Approved: yijinjing core now keeps only hash/mmap primitives and storage contracts; runtime-only stacktrace, terminal, signal, and AppContainer APIs moved to libkungfu runtime. Local check/build and embedding validation passed.

PR #407, opened 2026-07-08
refactor(core): retire longfist schema surface

kungfu-origin:
## Summary
- move the v4 runtime fact schema into `kungfu/yijinjing/schema` and remove the pre-v4 longfist public package from C++/Python/Node surfaces
- delete trading-era FlatBuffers schema/codegen and update the C++ probe to compile against yijinjing schema instead
- update docs and ADRs so v4 is the greenfield schema root, while v4 stable and later remain responsible for v4+ schema compatibility

## Verification
- `./kungfu-code check`
- `./kungfu-code build`
- `node framework/core/src/libyijinjing/check-deps.mjs`
- `node scripts/check-runtime-greenfield.mjs`
- `git diff --check`


dongkeren (approved):
Approved: schema surface cleanup is aligned with the v4 greenfield baseline and keeps v4+ compatibility responsibility explicit.

PR #434, opened 2026-07-09
docs(master): define supervisor topology

dongkeren:
## Summary
- add ADR-0036 for the per-user supervisor and per-data-root workspace master topology
- document that the supervisor routes and manages masters but does not own durable facts
- update master service, config, ADR index, and docs map entry points

## Validation
- git diff --check
- ./kungfu-code check

## Notes
- This is a documentation/architecture decision slice. Implementation work remains separate.


kungfu-origin (approved):
Reviewed architecture documentation split: per-user supervisor remains a process router, per-data-root master remains the Kungfu live coordinator, and durable storage stays daemonless.

PR #498, opened 2026-07-10
docs(kfx): propose native and wasm execution profiles

dongkeren:
## Summary

- document the historical and current native KFX evidence
- propose four orthogonal execution profiles: native, WebAssembly, managed, and subprocess
- recommend Rust as the native authoring default only behind a versioned C ABI and explicit safety gates
- select Wasmtime provisionally for a future component-model spike, with Wasmer retained as a measured fallback
- define performance, footprint, provenance, and decision gates without changing the current KFX contract

## Validation

- git diff --check
- local Markdown link validation
- staged repository gate passed during commit
- ./shifu check: 25/26 SDK tests passed; the remaining failure is a pre-existing canonical-policy hash mismatch in an unchanged contract file

## Scope

Research and proposed ADR only. No runtime, contract, dependency, or build-pipeline implementation is included.

### DIFF
--- a/framework/core/docs/adr/ADR-0045-kfx-execution-profiles-native-rust-wasm.md
+++ b/framework/core/docs/adr/ADR-0045-kfx-execution-profiles-native-rust-wasm.md
@@ -1,0 +1,280 @@
+---
+status: draft
+period: 2026-07-10
+theme: kfx-execution-profiles
+doc_type: architecture-decision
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Reviewed the proposed profile separation, evidence boundaries, implementation gates, and explicit non-implementation scope.

PR #500, opened 2026-07-10
docs(positioning): define the runtime fact infrastructure layer

dongkeren:
## Summary

- position Kungfu through the SQLite shape, Git-for-runs semantics, and flight-recorder-plus-qualification mission
- distinguish the runtime fact ledger from observability and blockchain
- explain why agent work creates a missing first-class runtime-fact layer
- route the positioning question and keywords from the documentation map

## Validation

- `git diff --check`
- public-surface leakage scan
- `./shifu check`


kungfu-origin (approved):
Reviewed: the positioning remains product-facing, distinguishes intent from current guarantees, and routes readers to the existing evidence boundaries.

PR #511, opened 2026-07-10
docs(architecture): establish dual schema authority

dongkeren:
## Summary

- establish ADR-0047: every persisted structured fact has exactly one schema owner, either Hana POD or FlatBuffers
- supersede ADR-0002 in schema scope and amend ADR-0025 so the action envelope migrates from transitional JSON/base64 to `ActionEnvelope.fbs`
- align public architecture, event, storage, carrier, and versioning docs; correct ADR-0037 so typed storage service work is no longer reported as complete

## Validation

- `./shifu check`
- `git diff --check`

## Version impact

- breaking-pre-release decision for the action-envelope encoding; no stable v4 compatibility promise is affected


kungfu-origin (approved):
Architecture decision and documentation are internally consistent. Validation is in progress; merge remains gated on required checks.

PR #550, opened 2026-07-11
feat(query): add SQLite and SQL conformance

dongkeren:
## Summary

- add an append-preserving SQLite Episode record projection that reuses the canonical decoder and fold
- execute one LogicalPlan through authority and SQLite engines with physical execution evidence
- add a fail-closed bounded SQL compiler plus CLI discovery, projection rebuild, and engine selection
- register the additive runtime-fact-query surface update and document accepted/rejected SQL

## Invariants

- the yijinjing Episode journal remains the sole authority
- SQL cannot own declaration roots, cut, time basis, or policy
- stale or absent projections fail closed
- authority and SQLite results must match at head, exact cuts, missing cuts, and rejected declaration bases

## Validation

- `./shifu build`
- `PYTHONPATH=src/python:build/Release uv run --frozen pytest -q tests/python/test_episode*.py tests/python/test_query_cli.py tests/python/test_atlas_storage.py` (92 passed)
- staged C++ format, Python format/lint, Shifu contract, action-envelope, and runtime greenfield gates

## Version impact

Additive update to the pre-release `runtime-fact-query` integration/cross-time surface. No existing field semantics or authority policy changed.

PR #551, opened 2026-07-11
feat(query): add SQLite SQL conformance

dongkeren:
Linear replacement for #550 after the repository rejected merge commits during rebase-only integration. Same validated tree as #550.\n\nValidation:\n- ./shifu build\n- 95 Python tests passed\n- staged gates passed\n\nNode binding suite discovered 7 tests but skipped all because the expected dist binding path was unavailable.


kungfu-origin (approved):
Q2 SQL conformance tree matches the validated #550 result and is linearized for the protected rebase-only branch.

PR #560, opened 2026-07-11
fix(storage): harden Hana SQLite projections

dongkeren:
## Summary

- preserve enum, fixed-array, and vector BLOB roundtrips through sqlite_orm
- rebuild Source, Manifest, and Episode projections atomically on one SQLite connection
- derive primary-key normalization and canonical content digests from Hana schema metadata
- keep verification read-only and detect same-row-count corruption, including append-only export subset integrity
- batch replay through sqlite_orm replace_range

## Mac validation

- `./shifu check`
- `./shifu verify --full` — 69/74; remaining failures are existing Episode contention / Atlas demo / rewind capture baselines
- targeted storage regression suite — 51 passed
- `storage-demo-node-binding` fixture passed
- Hana SQLite capability slice passed
- ASan/UBSan view replay passed

CI is not used as the merge gate for this change.


kungfu-origin (approved):
Reviewed the Hana schema ownership, single-connection transaction boundary, read-only verification path, and Mac validation evidence.

PR #581, opened 2026-07-11
fix(journal): harden mmap ownership and page publication

dongkeren:
## What changed

- introduce move-only `mapped_region` ownership for POSIX and Windows mappings
- separate existing-only mappings from explicit create/grow authority
- make `page_header::last_frame_position` the release/acquire page publication token
- validate page/header/frame sizes and offsets before payload access
- keep journal wire-v1 and the public three-argument Python `get_page_path` API unchanged
- extend ADR-0001 and add native mmap correctness tests wired through Shifu and CTest

## Why

Reader mappings could create or stretch files, raw mapping ownership leaked resources on error paths, and page initialization relied on ordinary cross-process field polling. The new boundary makes ownership and mutation authority explicit while preserving the existing single-writer, zero-copy journal contract.

## Validation

- `./shifu build:core` — passed on Mac arm64 after rebasing onto current `dev/v4/v4.0`; includes Node, Electron, Python, Wasmer and Wasmtime targets
- `./shifu test:mmap` — passed
- `ctest --test-dir framework/core/build -R '^yijinjing_mmap_tests$' --output-on-failure` — passed
- `./shifu check` — passed
- commit pre-checks including clang-format 20.1.8 — passed

CI is currently unavailable and is not used as the merge gate for this change; the requested gate is the Mac build and tests above.

### DIFF
--- a/framework/core/CMakeLists.txt
+++ b/framework/core/CMakeLists.txt
@@ -57,6 +57,10 @@ option(NNG_TOOLS "" OFF)
 option(NNG_ENABLE_NNGCAT "" OFF)
 option(KUNGFU_WITH_DEBUG_EXAMPLE "" OFF)
 option(KUNGFU_WITH_SLICES "build the capability slices (minimal standalone proofs, see slices/README.md)" OFF)
+option(KUNGFU_WITH_CORE_TESTS "build native yijinjing correctness tests" ON)
+if(KUNGFU_WITH_CORE_TESTS)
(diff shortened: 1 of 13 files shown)


kungfu-origin (approved):
Reviewed mmap ownership, existing-only reader authority, wire-v1 invariants, page publication ordering, bounds validation, and Mac validation evidence. Approved for dev.

PR #589, opened 2026-07-11
fix(api): tolerate runtime logs before Atlas JSON

dongkeren:
## Summary
- tolerate native runtime diagnostics before Atlas CLI JSON payloads
- keep malformed output fail-closed
- add regression coverage for the exact Work Dashboard failure shape

## Validation
- `./shifu --filter @kungfu-tech/api run test:query` (6/6)
- `./shifu check`
- `./shifu product gui build`
- `./shifu package` registered the fix build as Shifu build `[0]`

## Known unrelated gaps
- the live Atlas demo fixture currently fails earlier on `storage_manifest_invalid: sync_root_mismatch`
- CI is currently unhealthy; this PR is not waiting on CI for local dogfood verification

### DIFF
--- a/framework/api/package.json
+++ b/framework/api/package.json
@@ -19,7 +19,7 @@
   "main": "./src/capability/index.ts",
   "scripts": {
     "build": "tsc --noEmit",
-    "test:query": "node --test tests/query.test.ts tests/ledger.test.ts",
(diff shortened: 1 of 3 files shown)


dongkeren:
Closing as obsolete after #852 extracted Mission Control into the Profile domain capability and removed framework/api/src/capability/atlas.ts, which is the sole implementation target of this patch. The old branch no longer has a valid integration surface on the current development line.

PR #590, opened 2026-07-11
fix(ci): provision mirrored Rust and Cargo

### LINKED ISSUE #579: libwasm backends fail to build on the self-hosted matrix: stale cargo (linux) and no crates.io mirror (macos)
Run 29148449019 (dev/v4/v4.0, workflow_dispatch) reached the build via the native shifu lifecycle on all platforms; both failures are libwasm-specific runner-environment gaps, not code errors:

**Linux x64**: both backends fail immediately —
`error: failed to parse lock file at: .../crates/libwasm/wasmtime/Cargo.lock` (same for wasmer). The runner's cargo predates the lock-file format in use; the toolchain that builds `crates/libwasm` needs a minimum-version guarantee on self-hosted runners (the dev Macs were recently updated to stable 1.97; the linux runner was not).

**macOS ARM64**: wasmtime builds, wasmer fails in `bindgen v0.72.1` deps —
`failed to download from https://index.crates.io/re/ge/regex` (retries consumed ~30 min before dying). Self-hosted runners sit behind a constrained uplink; node/python/electron fetches all have mirror keys in build-local.env, but cargo has no registry mirror configured (~/.cargo/config.toml sparse-mirror or vendoring). Any crates.io-heavy build will be flaky until it gets one.

**Windows x64**: did not reach the cargo stage (unrelated lifecycle-shell gap, handled separately).

Suggested: pin/verify a minimum rust toolchain on self-hosted runners as part of the libwasm build contract, and add a cargo registry mirror key alongside the existing toolchain mirror set.

dongkeren:
Summary: provision Rust 1.96.0 on the self-hosted native matrix; use mirrored rustup distribution endpoints; pass the Cargo registry index through a repository variable so private LAN topology stays out of public workflow YAML. Validation: ./shifu check passed; full self-hosted workflow run 29156651239 is in progress. Depends on Buildchain #1117 and the stacked Cargo registry input PR. Fixes #579.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,11 +25,16 @@ permissions:
 
 jobs:
   build:
-    uses: kungfu-systems/buildchain/.github/workflows/build.yml@v2
+    uses: kungfu-systems/buildchain/.github/workflows/.build.yml@train/v2/v2.12/cargo-registry-mirror
     secrets: inherit
     with:
       buildchain-ref: ${{ inputs.buildchain-ref || '' }}
       runner-preset: kungfu-v4-self-hosted
+      setup-rust: true
+      rust-toolchain: "1.96.0"
+      rustup-dist-server: https://rsproxy.cn
+      rustup-update-root: https://rsproxy.cn/rustup
+      cargo-registry-index: ${{ vars.BUILDCHAIN_CARGO_REGISTRY_INDEX }}
       artifact-name: kungfu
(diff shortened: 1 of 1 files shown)


kungfu-origin (approved):
The provisioning candidate reached Cargo/libwasm on Windows, proving the Rust setup; its later mmap.cpp failure was independently fixed in #597. Re-run the combined tree through the alpha gate.

PR #597, opened 2026-07-11
fix(core): suppress Windows minmax macros in mmap

### LINKED ISSUE #596: Windows product build fails because windows.h max macro expands numeric_limits
The Rust/Cargo provisioning candidate exposed the next Windows release blocker in product build run 29156716599.

`framework/core/src/libyijinjing/src/util/mmap.cpp:91` includes `windows.h` without `NOMINMAX`, so `std::numeric_limits<LONGLONG>::max()` expands the Win32 `max` macro and MSVC reports C2589/C2059/C2143.

Use the repository's existing Windows convention: define `NOMINMAX` before `windows.h`, then rerun the self-hosted Windows product gate.

dongkeren:
Fixes #596.

The Windows product gate includes `windows.h` before using `std::numeric_limits<LONGLONG>::max()`. Defining `NOMINMAX` at the include boundary prevents Win32 macro expansion and follows the existing repository convention.

Evidence:
- exact MSVC failure reproduced by run 29156716599
- `./shifu check` passed

The follow-up self-hosted Windows product build will run together with the Rust/Cargo provisioning candidate.

### DIFF
--- a/framework/core/src/libyijinjing/src/util/mmap.cpp
+++ b/framework/core/src/libyijinjing/src/util/mmap.cpp
@@ -6,6 +6,9 @@
 
 #ifdef _WINDOWS
 #include <fcntl.h>
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
 #include <windows.h>
 #else
 


kungfu-origin (approved):
The MSVC failure is directly attributable to windows.h max macro expansion; NOMINMAX is the minimal repository-standard fix.

PR #599, opened 2026-07-11
perf(yijinjing): qualify and optimize mmap page lookup

dongkeren:
## Summary

- add a reproducible native mmap qualification harness and retain macOS/Linux evidence
- replace reverse-linear page-header lookup with a tail fast path plus ordered binary probes
- record independent accept/reject/defer decisions for sizing, advice, residency, flush range, and release polling

## Evidence

- Mac three-round A/B at 128 pages: early seek p50 3.131-3.505 ms -> 0.298-0.517 ms; middle 1.825-1.953 ms -> 0.293-0.454 ms; late unchanged
- Linux three-round A/B: early 0.750-0.782 ms -> 0.153-0.155 ms; middle 0.431-0.454 ms -> 0.152-0.154 ms; late unchanged
- `./shifu rebuild` passed on macOS arm64 and agent-120 Linux x86_64
- `./shifu test:mmap` passed on both hosts
- `./shifu check` passed on macOS

## Compatibility

No wire bytes, Hana POD layout, public API, mapping authority, or durability semantics change. Rollback is a normal revert of the two lookup commits.


kungfu-origin (approved):
Reviewed qualification evidence, compatibility boundary, and Mac/Linux validation. Ordered lookup is isolated and rollback-safe.

PR #605, opened 2026-07-11
fix(ci): align libwasm Rust toolchain pins

### LINKED ISSUE #604: libwasm toolchain pins race the provisioned Rust version in parallel builds
Alpha r2 run 29157376683 provisions Rust 1.96.0 through the Buildchain runner contract, but both `crates/libwasm/rust-toolchain.toml` and `crates/libwasm-spike/rust-toolchain.toml` still pin 1.95.0.

On Linux, parallel Wasmer/Wasmtime CMake targets therefore launched two rustup installs of 1.95.0 against the same rustup home. One failed renaming a `.partial` component; the other recovered the partial install and removed cargo, then failed because cargo was not applicable.

Align the repository toolchain pins and documentation to the provisioned 1.96.0 contract, then rerun the three-platform alpha gate.

dongkeren:
Fixes #604.

Align both libwasm toolchain pins with the Rust 1.96.0 version provisioned by the self-hosted Buildchain runner contract. This prevents parallel Wasmer/Wasmtime targets from racing two implicit rustup installs of the old 1.95.0 toolchain.

Evidence:
- exact race captured in alpha r2 Linux run 29157376683
- `./shifu check` passed

A replacement alpha candidate will rerun all three product legs after merge.

### DIFF
--- a/crates/libwasm-spike/rust-toolchain.toml
+++ b/crates/libwasm-spike/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.95.0"
+channel = "1.96.0"
 profile = "minimal"
--- a/crates/libwasm/rust-toolchain.toml
+++ b/crates/libwasm/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.95.0"
+channel = "1.96.0"
 profile = "minimal"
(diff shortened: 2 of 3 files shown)


kungfu-origin (approved):
Rust 1.96 pin alignment is minimal and the immutable three-platform gate passes, including the Windows rerun with native-kfx aggregate p99 4900ns under the unchanged 5us budget.

PR #629, opened 2026-07-12
release: promote current v4.0 dev to alpha

dongkeren:
## Summary

- promote the complete current `dev/v4/v4.0` tree at `aff6e13da` to alpha
- retain Buildchain KFD onboarding, Shifu pin-first acquisition, and the cross-platform product-gate repairs
- include the merged C++23/RxCpp portability contract from #627
- preserve alpha history while making the candidate tree byte-identical to the locked dev tree

## Validation

- candidate tree matches `aff6e13da` exactly
- staged repository gate passed during the merge commit, including Rust tests and Buildchain KFD evidence
- #627 carries three-platform local qualification for the native toolchain contract
- this PR product matrix is the release authority

Supersedes #621.

PR #632, opened 2026-07-12
fix(product): select ESM entrypoint specifiers by platform

dongkeren:
## Summary

- keep absolute filesystem paths for POSIX first-party manifest generation
- use a `file://` URL only on Windows, where drive-letter paths require it
- cover Linux, macOS, and Windows specifier selection

## Validation

- `./shifu install`
- `node --test product/scripts/dist.test.mjs`

Fixes the Linux product failure in #629.


kungfu-origin (approved):
Verified the platform-specific ESM entrypoint handling and focused regression coverage.

PR #640, opened 2026-07-12
fix(gui): default desktop packaging to no publish

dongkeren:
## Summary
- default the cross-platform electron-builder launcher to `--publish=never`
- preserve an explicit caller-provided publish mode
- make the launcher import-safe and cover both argument paths with focused tests

## Context
Kungfu r6 generated the Linux AppImage successfully, then electron-builder inferred CI publishing and failed because candidate builds intentionally do not receive a GitHub publishing token. Release publication remains owned by Buildchain.

## Validation
- `node --test framework/gui/scripts/run-electron-builder.test.mjs product/scripts/dist.test.mjs` (5 passed)
- staged repository gate passed
- `git diff --check`
- failing product evidence: run 29186140338


kungfu-origin (approved):
Verified the no-publish default, explicit override preservation, focused tests, and current dev integration.

PR #641, opened 2026-07-12
fix(ci): align membrane native toolchain

dongkeren:
## Summary
- select Ninja for the explicit membrane consumer CMake configure
- run the Linux qualification lane on Ubuntu 24.04 with GCC 14
- align the spike workflow with the repository native toolchain contract

## Context
The native toolchain contract requires Ninja and GCC 14 or newer. The membrane workflow previously defaulted to Unix Makefiles on Ubuntu 22.04/GCC 11 and failed before exercising consumers.

## Validation
- staged repository gate passed
- `git diff --check`
- Ninja mismatch reproduced in #640 run 29186472119
- GCC mismatch reproduced in this PR run 29186603764

### DIFF
--- a/.github/workflows/embedding-membrane-spike.yml
+++ b/.github/workflows/embedding-membrane-spike.yml
@@ -77,7 +77,7 @@ jobs:
       matrix:
         include:
           - name: Linux x64
-            runner: '"ubuntu-22.04"'
+            runner: '"ubuntu-24.04"'
             source_mode: github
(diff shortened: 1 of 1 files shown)


dongkeren:
Superseded by the narrower #640 launcher fix. This PR also exposed separate membrane qualification debt (hosted macOS RocksDB/C++23 and Windows compiler environment), which is outside the Buildchain artifact-onboarding release blocker.

PR #643, opened 2026-07-12
perf(libwasm): share stable per-engine Cargo caches

dongkeren:
## What changed

- move production Wasmtime and Wasmer Cargo target directories out of each
  CMake build tree into stable per-engine user-cache slots;
- key each slot by source checkout, actual Cargo/rustc identity and target,
  profile, engine, and lockfile;
- stage each adapter back into its consuming CMake tree so build artifacts do
  not depend on an external-cache path;
- serialize the two heavy Cargo builds inside one CMake graph while retaining
  Cargo's per-target lock as the cross-graph boundary;
- apply the same contract to the shared-membrane qualification slice;
- add a deterministic CMake contract test and
  `./shifu qualify:libwasm-cache` for cold/warm cross-platform evidence.

## Why

The two engine workspaces intentionally remain separate because their Cranelift
dependency lines are incompatible. The previous integration also placed both
target directories under `CMAKE_CURRENT_BINARY_DIR`, so every distinct CMake
tree paid another full Rust dependency build. This change preserves the strict
engine/workspace boundary while making target reuse stable and explicit.

## Impact

Separate CMake trees from one checkout now reuse compiled Rust dependencies.
Different worktrees, engines, toolchains, targets, profiles, or lockfiles cannot
collide. `CARGO_HOME` still owns shared registry downloads; the optional Cargo
mirror remains a transport-only override.

## Validation

- macOS arm64:
  - qualification (Cargo 1.95): Wasmtime `319 ms -> 89 ms`, Wasmer
    `327 ms -> 132 ms` on the corrected warm run;
  - full `./shifu build:core` passed in `117.50 s`; production CMake adapter
    steps reused the cache in `0.45 s` and `0.66 s`.
- agent-120 Linux x64 (Cargo 1.96): Wasmtime `80.2 s -> 78 ms`, Wasmer
  `76.7 s -> 91 ms`.
- DARKHERO Windows x64 (Cargo 1.96): Wasmtime `159.3 s -> 112 ms`, Wasmer
  `164.7 s -> 202 ms`.
- Commit hooks: staged source/format/contract gates passed on both commits.
- `./shifu check`: all changed-file and cache-contract gates passed; the shared
  tooling phase remains blocked by the pre-existing unchanged
  `framework/core/.gyp/run-freeze.js:600` TypeScript inference error.
- CI was not used for this change; the three requested native hosts supplied
  the build evidence.

### DIFF
--- a/build-local.env.example
+++ b/build-local.env.example
@@ -56,6 +56,12 @@
 # thumb: jobs ≈ available memory GB / 2. If unset, conan defaults (os.cpu_count).
 # export KUNGFU_BUILD_JOBS="12"
 
+# Optional libwasm Cargo target-cache root. By default every checkout uses a
+# deterministic per-engine slot below ${XDG_CACHE_HOME:-$HOME/.cache}/kungfu,
+# shared by all of that checkout's CMake build trees. Set this only when a
+# managed runner wants the cache on a dedicated local volume.
+# export KF_LIBWASM_CARGO_TARGET_ROOT="/path/to/local-cache/kungfu/libwasm/cargo-target"
+
 # ── Dev build stash (shifu promote / builds) ────────────────────────────────
 # Successful desktop builds register themselves user-globally
 # (~/.cache/kungfu/product/<os>-<arch>/) so `shifu promote` can install the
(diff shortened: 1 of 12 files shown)


kungfu-origin (approved):
Reviewed cache-key boundaries, per-engine isolation, toolchain cwd alignment, and three-host qualification evidence. macOS build:core and macOS/Linux/Windows cold-warm qualification provide sufficient merge evidence while CI is unavailable.

PR #656, opened 2026-07-12

dongkeren:
## Summary
- verify product Buildchain log events only from the current process window
- ignore stale same-component errors retained by a self-hosted workspace across reruns without deleting the shared JSONL
- fix the NodeNext `.js` extension missing from the newly added Atlas saved-query type import

## Validation
- `./shifu check`
- product observability regression covers prior-run `kungfu-product` errors
- tooling type check covers the NodeNext import

Kungfu r8 direct qualification already proved AppImage creation and the installed CLI layout smoke; it then failed only because 32 error events from prior attempts remained in the append-only log.

### DIFF
--- a/framework/api/src/capability/atlas.ts
+++ b/framework/api/src/capability/atlas.ts
@@ -1,4 +1,4 @@
-import type { SavedQueryView } from './query';
+import type { SavedQueryView } from './query.js';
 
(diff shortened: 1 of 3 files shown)


dongkeren:
Superseded by #688: a clean one-commit branch rebased on current dev. This avoids staging unrelated dev history and preserves the same observability regression test and implementation.

PR #659, opened 2026-07-12
fix(ci): qualify membranes on current toolchains

dongkeren:
## Summary
- run membrane qualification through the pinned Shifu entrypoint on every platform
- select Ubuntu 24.04 with GCC 14 and Ninja, matching the checked-in toolchain contract
- export an audited RocksDB 6.29.5 compatibility recipe with current AppleClang/C++23 fixes while preserving the persisted-data version contract
- let Shifu materialize the MSVC environment on Windows
- include the NodeNext Atlas import fix that currently blocks the repository type gate

## Validation
- `./shifu check`
- prior direct failures reproduced all three gaps: Linux generator/compiler, macOS RocksDB incomplete types, Windows compiler environment
- this PR must pass the three-platform `Embedding membrane spikes` matrix before merge

### DIFF
--- a/.github/workflows/embedding-membrane-spike.yml
+++ b/.github/workflows/embedding-membrane-spike.yml
@@ -213,7 +213,7 @@ jobs:
             spike_python="$PWD/framework/core/.venv/Scripts/python.exe"
           fi
           # shifu-entry-contract: allow the isolated spike configure is not a product-level Shifu task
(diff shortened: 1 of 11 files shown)


dongkeren:
Superseded by #662 after #661 entered dev. The replacement branch was rebuilt linearly on current dev without force-pushing this published branch.

PR #661, opened 2026-07-12
fix(api): use NodeNext query specifier

### LINKED ISSUE #658: Fix NodeNext extension for Atlas capability import
Current dev fails ./shifu check in the tooling type check:

framework/api/src/capability/atlas.ts imports ./query without the explicit .js extension required by NodeNext module resolution.

Acceptance criteria:
- use the runtime .js specifier for the type-only relative import
- ./shifu check:types passes
- ./shifu check passes

dongkeren:
Closes #658.

Use an explicit `.js` specifier for the type-only `query` import so the API package remains valid under NodeNext module resolution.

Validation:
- `./shifu check:types`
- `./shifu check`
- `git diff --check`

### DIFF
--- a/framework/api/src/capability/atlas.ts
+++ b/framework/api/src/capability/atlas.ts
@@ -1,4 +1,4 @@
-import type { SavedQueryView } from './query';
+import type { SavedQueryView } from './query.js';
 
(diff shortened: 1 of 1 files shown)


kungfu-origin (approved):
NodeNext specifier fix is minimal and the type/full checks are green.

PR #670, opened 2026-07-12
chore(ci): drop dead source-mode mirror config and orphan checks.json

dongkeren:
Dead-config cleanup (zero behavior change), from a dead-code survey.

- **`embedding-membrane-spike.yml`**: the matrix only ever sets
  `source_mode: github`, so the `if [ "$SOURCE_MODE" = "mirror" ]` reconstruct
  branch never runs and its `MIRROR_REF` / `MIRROR_URL` env are unreachable —
  removed. The live github reconstruct path and `MIRROR_BASE_SHA` (still used)
  are unchanged; the full native matrix on this PR revalidates the step.
- **`checks.json`** (repo root, `{"enabled":true,"categories":{}}`, dated 2023):
  no reference anywhere in the repo — removed.

Scope is intentionally minimal: the vestigial always-true `source_mode == 'github'`
guards are left in place (a larger simplification is tracked separately), and
all code-level dead-code candidates are deferred to a review card.

### DIFF
--- a/.github/workflows/embedding-membrane-spike.yml
+++ b/.github/workflows/embedding-membrane-spike.yml
@@ -102,8 +102,6 @@ jobs:
       CARGO_HTTP_TIMEOUT: "120"
       CARGO_NET_RETRY: "5"
       MIRROR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
-      MIRROR_REF: refs/heads/dev/v4/v4.0
-      MIRROR_URL: http://192.168.100.222:8088/git-mirrors/kungfu.git
       SOURCE_MODE: ${{ matrix.source_mode }}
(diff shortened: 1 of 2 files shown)


dongkeren:
Parking: the 3 native jobs fail at **Generate core toolchain** (rocksdb/6.29.5 fails to compile against Xcode 16.4 / MacOSX15.5 libc++ — `incomplete type ParallelCompressionRep`), which is a repo-wide environmental toolchain breakage already being fixed on separate `fix/membrane-*toolchain*` branches — unrelated to this diff. This PR's own edited step (`Reconstruct and verify immutable PR head`) passed on all 3 platforms. Will rebase + re-run once the toolchain fix lands on dev/v4/v4.0, then merge.

kungfu-origin (approved):

PR #675, opened 2026-07-12
fix(gui): keep Mission Control refresh nonblocking

dongkeren:
## Summary

- execute Mission Control dashboard and assessment CLI reads asynchronously in the Electron main process
- apply one cut-bound dashboard snapshot atomically with cached stale-while-refresh behavior, single-flight coalescing, and stale result rejection
- yield between query changelog pages and remove synchronous native runtime-health polling from the renderer

## Validation

- `./shifu check`
- API query tests, GUI Atlas transport tests, dashboard refresh coordinator tests
- focused Python Mission Control snapshot test
- full core build and `./shifu dist --product desktop --dir`
- real `/Users/dkr/Code/atlas` 10-second continuous resize with 12 concurrent refreshes: RAF p95 17.6 ms, max 17.8 ms, zero long tasks, no degraded snapshot

CI is currently known to be unreliable; this change is being assessed from the local evidence above.


kungfu-origin (approved):
Reviewed the asynchronous boundary, stale-result protection, renderer polling removal, and local performance evidence.

PR #688, opened 2026-07-12

dongkeren:
## Summary
- snapshot the Buildchain event count when the product process starts
- verify only events emitted by the current process so append-only logs from prior reruns cannot poison release qualification
- retain component filtering and add a regression fixture for stale same-component errors

## Validation
- `./shifu check`

Supersedes #656 with a clean one-commit branch on current dev.


dongkeren:
Superseded by clean current-dev replacement #719.

PR #689, opened 2026-07-12
style(schema): format manifest layout weld

dongkeren:
Format the ADR-0067 manifest layout weld macro with the repository-pinned clang-format.

This was exposed by the alpha candidate staged gate, which correctly checks the entire dev projection against the older alpha base. No behavior changes.

Validation:
- `uvx clang-format@20.1.8 -style=file --dry-run -Werror framework/core/src/libyijinjing/include/kungfu/yijinjing/schema/registry.h`
- `./shifu check`
- `git diff --check`


kungfu-origin (approved):
Pinned clang-format produces the only diff; checks are green and behavior is unchanged.

PR #691, opened 2026-07-12
feat(runtime): persist durable frame context

dongkeren:
## Summary
- supersede the internal test-only KFDL v1 container with v2 records carrying the complete journal frame context
- expose the context through durable records and the state-service shadow append boundary
- verify restart reconstruction against an actually published mmap frame without changing journal wire-v1 or the mmap hot path

## Why
The projection bootstrap shadow could prove payload identity, but existing `state<DataType>` routing also requires source, destination, generation time, data type, and trigger lineage. Persisting those facts in the independent durable record is required before a real typed-state projector can be qualified.

## Validation
- `./shifu build:core`
- `./shifu test:durable-ingest`
- `./shifu test:projection-bootstrap`
- `./shifu test:state-service`
- `./shifu test:durability-contract`
- `./shifu test:runtime-errors`
- `./shifu test:mmap`
- `./shifu check`


kungfu-origin (approved):
Reviewed the KFDL v2 frame-context extension, restart integrity checks, unchanged mmap wire/hot path, and fail-closed projection prerequisite.

PR #697, opened 2026-07-12
feat(runtime): hydrate typed state across restart

dongkeren:
## Summary
- decode verified typed-state images into a staging `state_cache::bank` and atomically replace the target only after full validation
- preserve the target bank on truncated, unknown-type, uid-mismatched, or malformed image data
- add a two-process fixture: creator writes a qualified test KFDL cut and snapshot; verifier reopens both, bootstraps, and hydrates typed peer state

## Boundary
This completes the process-independent hydration primitive, not the production coordinator cutover. The compatibility restore remains active until a verified production profile and authority wiring are available; journal wire-v1 and the mmap hot path are unchanged.

## Validation
- `./shifu build:core`
- `./shifu test:projection-bootstrap`
- `./shifu test:durable-ingest`
- `./shifu test:state-service`
- `./shifu test:durability-contract`
- `./shifu test:runtime-errors`
- `./shifu test:mmap`
- `./shifu check`


kungfu-origin (approved):
Reviewed atomic staged hydration, target rollback on decode failure, the two-process KFDL/snapshot restart chain, and the explicit no-production-cutover boundary.

PR #708, opened 2026-07-12
fix(docs): keep executable example source-only

dongkeren:
The executable Shifu version example now forces the source-only script path. This removes native launcher compilation from the bounded docs example and fixes the GitHub-hosted docs gate timeout observed after the publication-operations merge.


kungfu-origin (approved):
Verified the source-only example path and local docs contract suite.

PR #719, opened 2026-07-12

dongkeren:
## Summary

- snapshot the Buildchain log-event index when the product process starts
- verify only current-process events instead of re-reading stale same-component failures from prior runs
- keep sibling-component filtering and add a regression fixture for stale `kungfu-product` errors

## Validation

- `./shifu check`
- regression test: `product observability ignores same-component errors from prior runs`


dongkeren:
Closing this line of work. The ad hoc hosted-registry split is superseded by a unified cache-management design at the Shifu layer. Branches and commits are retained for audit/recovery; no release artifacts are deleted.

PR #729, opened 2026-07-12
ci(core): use upstream Cargo index on hosted runners

dongkeren:
## Summary

- make the libwasm qualification Cargo registry configurable while retaining rsproxy as the local/default path
- route GitHub-hosted membrane qualification through the official sparse crates.io index
- keep the existing 120-second timeout and five retry settings as transport resilience

## Validation

- `./shifu check`
- run 29213431986 Linux showed rsproxy was completely unreachable for the full 120-second timeout; this change removes that hosted-only dependency without changing local mirror policy

## Delivery intent

This PR restores deterministic hosted-runner qualification without changing the local mirror policy or product runtime behavior.


kungfu-origin (approved):
Reviewed the hosted-runner Cargo source selection. An empty override correctly preserves Cargo built-in crates.io while local defaults retain the mirror.

dongkeren:
Closing this line of work. The ad hoc hosted-registry split is superseded by a unified cache-management design at the Shifu layer. Branches and commits are retained for audit/recovery; no release artifacts are deleted.

PR #780, opened 2026-07-13
fix(profile): bind declared home view

dongkeren:
## Summary

- accept the KFX `experience.homeView` field in the libkungfu Profile authority
- require the declared home view to resolve to a Suite member
- preserve the field in the normalized, content-bound Profile root

## Verification

- `./shifu rebuild:core`
- native Profile lifecycle tests (Node and Python build variants)
- `./shifu freeze && node framework/gui/scripts/gen-system-profile-kfd3.mjs`
- `./shifu check:source`

## Governance risk check

- [x] none of the above

## Checklist


kungfu-origin (approved):
Approved: Profile authority now matches the shared KFX contract and fail-closes on an unbound home view.

PR #784, opened 2026-07-13
fix(core): preserve RocksDB source archive type

dongkeren:
## Summary

Give the RocksDB Conan source download an explicit `.tar.gz` filename. The codeload URL ends in a commit hash, so Conan otherwise saves the valid gzip archive without an extension and attempts ZIP extraction, producing the same `BadZipFile` failure on macOS, Linux, and Windows.

## Changes

- retain the existing source URL, SHA-256 verification, and `strip_root` behavior
- name the temporary source object `rocksdb-source.tar.gz`
- add a build-free regression assertion to the source acceptance suite

## Verification

- no-controller-cache `./shifu check:source` passed with 101 source-contract tests
- pre-fix local Shifu `build:core` reproduced `BadZipFile: File is not a zip file` on macOS, Linux, and Windows
- post-fix local no-controller-cache Shifu `build:core` passed at `1140d28d9f5e2ff17a22596d0fb71a8f1398c7bf` on:
  - macOS arm64 / Apple Clang 21 / C++23
  - Linux x86_64 / GCC 14 / C++23
  - Windows x86_64 / MSVC 19.51 / C++23
- all three builds used the existing host Conan cache through the normal Shifu command; no global cache configuration was changed

## Governance risk check

- [ ] none of the above

The evidence remains local-only and revision-bound. This bugfix does not change workflows, Buildchain configuration, Shifu protocol, cache ownership, source URL, source digest, release artifacts, or the declared durability envelope.

## Checklist

### DIFF
--- a/docs/adr/ADR-0068-tiered-durability-and-crash-recovery.md
+++ b/docs/adr/ADR-0068-tiered-durability-and-crash-recovery.md
@@ -4,7 +4,7 @@ doc_type: architecture-decision
 adr_id: ADR-0068
 decision_status: accepted
 implementation_status: staged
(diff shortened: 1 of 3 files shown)


dongkeren:
Superseded by #788, which replays the same validated source fix and ADR projection directly on the latest protected dev head after the local up-to-date merge exposed the existing stale KFD witness. No force push or gate bypass was used.

PR #801, opened 2026-07-14
fix(core): sync generated peer runtime stubs

dongkeren:
## Summary

- sync the committed Python runtime stub with the five peer methods already exposed by the native binding
- restore clean Product provenance after deterministic core rebuilds

## Validation

- generated diff reproduced identically across two full macOS Product builds
- staged repository gates passed
- prior exact merged-head Product build completed with verified KFD-3 receipt


kungfu-origin (approved):
Generated projection matches the native peer surface reproduced by the macOS Product build; scoped and ADR-neutral.

PR #856, opened 2026-07-14
feat(agent-session): detach product capsule worker

dongkeren:
## Summary

- detach Agent Session Capsule ownership from Electron main into one private local worker with a stable, runtime-scoped endpoint and reconnectable RPC client
- serialize missing-worker startup across independent clients with a private atomic lock so no client can unlink another workers live socket
- fence shutdown across the worker, RPC surface, runtime, and provider PTYs; a missing worker starts a new empty runtime instead of claiming recovery
- harden bounded VT/provider state classification against stale screens while keeping volatile terminal text memory-only
- retain machine-readable Stage 6 qualification evidence and a metadata-only real-provider dogfood runner

## Verification

- `./shifu test:agent-session-recovery-qualification` (real node-pty; retained p95 1.645 ms, final rerun p95 4.402 ms, ceiling 250 ms)
- `./shifu check:source` after review fix (246/246 source tests; includes independent-client startup race regression)
- `./shifu --filter @kungfu-tech/gui build` after rebase (main, preload, renderer passed)
- authenticated Codex 0.144.3: instruction/output, approval deny, interrupt, Electron-main reconnect, provider termination all passed

## Known limits

- authenticated Claude 2.1.209 reached ready, instruction/output, and main-process reconnect, but its tool-approval surface did not converge within the bounded qualification window
- Claude deny input was therefore not sent and no Claude approval outcome is claimed; the attempted side effect did not execute
- evidence records `promotionEligible=false`; this PR does not build, register, or promote a Kungfu product artifact
- machine restart durability remains unqualified; worker loss intentionally produces a new empty runtime and attempt
- the package-wide native peer test requires a prebuilt Core binding; the source-only PR gate and scoped Stage 6 qualification do not claim that native build

## Governance risk check

- [x] none of the above

## Checklist


kungfu-origin (approved):
Independent review passed after adding cross-client worker-start serialization and the startup-lock release race regression coverage. The detached-worker implementation, bounded recovery/privacy evidence, ADR-0081 partial projection, and nine required CI checks are coherent. This approval covers the implementation slice only: authenticated Claude approval/deny, machine restart, Linux/Windows product qualification, and promoted Mac product evidence remain explicit blockers; promotionEligible stays false.

PR #913, opened 2026-07-15
fix(gui): skip Python bytecode during macOS signing

dongkeren:
## Summary

- exclude Python bytecode and `__pycache__` contents from the macOS code-signing walk
- preserve every existing ignore rule and continue signing native libraries and executables
- avoid thousands of unnecessary Apple timestamp requests that made signed product packaging intermittently fail

## Failure evidence

Two exact-source macOS qualification attempts reached `@electron/osx-sign` and failed on different `.pyc` files with `A timestamp was expected but was not found`; signing the same failed file manually succeeded immediately afterward. The failure is therefore isolated to high-volume timestamping of non-code Python cache artifacts.

## Validation

- `node --test framework/gui/scripts/sign-macos.test.mjs` (5 passed)
- `./shifu check:source` (316 source contract, 76 runtime upgrade, 66 desktop adapter tests passed)
- commit staged gate passed

### DIFF
--- a/framework/gui/scripts/sign-macos.mjs
+++ b/framework/gui/scripts/sign-macos.mjs
@@ -3,6 +3,7 @@
 import { signAsync } from '@electron/osx-sign';
 
 const CERTIFICATE_HASH = /^[A-F0-9]{40}$/i;
+const PYTHON_BYTECODE = /(?:^|[\\/])__pycache__(?:[\\/]|$)|\.py[co]$/i;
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Reviewed the bounded macOS signing exclusion: Python bytecode is excluded while native libraries, executables, and existing ignore rules remain covered. Source and staged gates are green.

PR #916, opened 2026-07-15
fix(gui): preserve macOS signing ignore function

dongkeren:
## Summary

- pass one combined ignore function to `@electron/osx-sign` instead of an array
- preserve existing string, array, and function rules inside that function
- keep Python bytecode out of code signing after the previous array form was discarded by osx-sign 1.3.3

## Runtime evidence

An exact-source signed product build after PR #913 still spawned `codesign` for a `.pyc` file. Inspection showed osx-sign 1.3.3 normalizes a non-array ignore into an array but returns `undefined` for an array input, silently discarding the prior hook result. The invalid signing run was stopped before further timestamp load.

## Validation

- `node --test framework/gui/scripts/sign-macos.test.mjs` (6 passed)
- `./shifu check:source` (316 source contract, 76 runtime upgrade, 67 desktop adapter tests passed)
- commit staged gate passed

### DIFF
--- a/framework/gui/scripts/sign-macos.mjs
+++ b/framework/gui/scripts/sign-macos.mjs
@@ -13,8 +13,14 @@ export function resolveMacSigningIdentity(options, env = process.env) {
 }
 
 export function resolveMacSigningIgnore(ignore) {
-  const configured = ignore === undefined ? [] : [ignore].flat();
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Reviewed against the observed osx-sign 1.3.3 normalization behavior. The adapter now supplies one combined function, preserves prior rules, and retains native binary signing.

PR #920, opened 2026-07-15
fix(ci): watch the membrane wrapper and its contract test in the membrane gate

dongkeren:
## Summary

Two honesty gaps left by #904, found while checking whether the cross-membrane
symmetry had actually been qualified on all three platforms. It had not, and the
reason turned out to be structural rather than a matter of waiting.

## Related issue

Follow-up to #904 (ADR-0078 cross-membrane decode symmetry).

## Changes

- **The membrane gate did not watch the membrane.** `embedding.membranes` is the
  gate that qualifies the embedding membrane, but its path filter omitted three
  files that decide whether the membrane contract still holds: the Rust wrapper
  crate (`crates/kungfu-embedding/**`), the generic-codec contract test itself,
  and the `CMakeLists.txt` that wires that test into ctest. A change confined to
  any of them would land without the gate ever running — the gate could not see
  the thing it exists to guard. #904 only triggered it (prospectively) because it
  happened to also touch `embedding.cpp`. Added all three to the filter.
- **ADR-0078 implied more evidence than exists.** The follow-up said the contract
  test "is wired into the embedding-membrane qualification gate, which runs on
  Linux, macOS, and Windows", and that Linux/Windows evidence "comes from the
  qualification gate". Both are true about wiring and read as though the evidence
  had been produced. It has not: the gate fires only on `alpha/*` and `release/*`
  pull requests — heavy native membrane builds are deliberately kept out of
  `dev/*` PRs (c727d5fa8) — so it has never run for this change and will not
  before the next `dev` → `alpha` promotion. The wording now says that plainly.

`implementation_status` stays `staged`; this PR does not change what has been
qualified, only what the repository claims and what the gate watches.

## Verification

- The workflow YAML parses, and the three previously unwatched paths now match
  the filter (checked against the parsed `pull_request.paths` list rather than by
  eye).
- `adr-audit` passes (`result=pass`, structural=0); the deterministic docs gate
  passes.
- No runtime surface is touched, so there is nothing to build or exercise here.


kungfu-origin (approved):

dongkeren:
Superseded by the current-dev linear replacement #1044, which preserved the membrane gate/ADR correction, refreshed workflow authority, passed all required branch and merge-group checks, and merged as cc89df26bd9fb7bb0f22671d53a582b185f9ee92.

PR #964, opened 2026-07-15
fix(core): probe Windows liveness without signaling

dongkeren:
## Summary

- probe Windows lock-holder liveness through `OpenProcess` / `GetExitCodeProcess` instead of `os.kill(pid, 0)`
- use psutil's non-signaling PID query for runtime service adoption checks
- anchor contract registry discovery at the module directory and make interrupted Windows table guards close without unlocking a lock they never acquired
- retain the welded Windows `call shifu.cmd` qualification surface

## Root cause and evidence

- hosted runs `29416494142`, `29420852660`, and self-hosted run `29422161962` reproduced `Terminate batch job` during the concurrent `activation-core` case
- verbose run `29426977705` exposed one independent Windows source-discovery defect, fixed here by starting at `contract.py`'s parent
- exact-topology run `29429837068` then exposed the decisive causal chain: the second activation caller invokes `_pid_alive(holder_pid)`; on Windows `os.kill(pid, 0)` is `CTRL_C_EVENT`, so the liveness probe interrupts every process sharing the console, including pytest and the outer batch
- the same run showed expected lock contention (`PermissionError` from `LK_NBLCK`) immediately before the Ctrl-C cascade; the lock was an effect/observer, not cross-thread business coupling

## Validation

- focused contract, coordination lock, runtime broker, and runtime service suite: 72 passed
- Windows-like regressions assert PID liveness probes never call `os.kill(pid, 0)`
- Ruff format/lint, `git diff --check`, and staged source acceptance hook passed

## Governance risk check

- [ ] none of the above

This PR repairs qualification and runtime liveness observation only; it does not publish or deploy artifacts.

## Checklist

- [x] Source acceptance is green locally
- [x] No credentials or generated qualification evidence are committed

### DIFF
--- a/framework/core/src/python/kungfu/contract.py
+++ b/framework/core/src/python/kungfu/contract.py
@@ -35,7 +35,7 @@ def resolve_registry_path(
         if product_candidate.is_file():
             return str(product_candidate)
 
-    for start in [Path(__file__).resolve(), Path.cwd().resolve()]:
+    for start in [Path(__file__).resolve().parent, Path.cwd().resolve()]:
         for directory in [start, *start.parents]:
             for rel in [
                 Path("framework") / "contract" / REGISTRY_FILE,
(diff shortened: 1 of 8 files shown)


kungfu-origin (approved):
Windows hosted proof and required profile checks establish the liveness-probe fix; approving for rebase merge once all required checks pass.

PR #988, opened 2026-07-16
refactor(query): parse bounded SQL with a tokenizer and recursive descent

dongkeren:
The MATCH_RECOGNIZE dialect was matched by a single ~700 character regular expression, with a second one for the plain episode form. Every optional clause meant another capture group to renumber, no clause could be reused, and a rejection could only report `unsupported SQL` for the whole statement.

This replaces both with a tokenizer and a recursive-descent parser, one function per clause, and reports the byte offset of the offending token while keeping the accepted-forms hint as a suffix.

## Equivalence

The accepted language is unchanged. An offline differential harness compared the regexes against the parser over a hand-written corpus plus byte-level mutation fuzz — **197901 inputs, no disagreement**. The harness was validated by confirming it catches six injected faults, including the exact 64/65 field-name and 256/257 value-length boundaries, so the agreement is evidence rather than a vacuous pass.

Two details the regexes encoded implicitly and the parser now states outright:

- tokens carry whether whitespace preceded them, because the dialect requires a separator at some junctions (`SELECT *`) and allows none at others (`title='x'`);
- the identifier classes accept upper case, because both regexes were case-insensitive and their captures were folded afterwards.

## Tests

A corpus test pins the accepted language now that there is no second implementation to compare against. Accepted statements assert every captured field, not merely that parsing succeeded.

Verified locally: full core build; 19/19 native tests; 11/11 query CLI tests; the SQL frontend storage tests pass.

Note: `test_sql_frontend_and_sqlite_projection_conform_at_head_and_exact_cut` fails on the current dev baseline, unrelated to this change — it asserts two `execution_engines` while `query_capabilities_json` has returned three since 9da49c3688. Left for a separate fix.

## Risk

This parser is the input trust boundary. The quoted-value character class and both length bounds are carried over unchanged and are covered by the corpus negatives, including quote, escape and separator injection.


kungfu-origin (approved):

dongkeren:
Superseded by the current-dev linear replacement #1045, which preserved the parser implementation, corrected affected-native architecture ownership, passed all required checks, and merged through the dev merge queue as 4ea73c1a028ce708fd47c44f3b3b7f6d7efc4490.

PR #990, opened 2026-07-16
refactor(yijinjing): clear journal-layer fossils and bound frame payload writes

kungfu-origin:
## What

Three related cleanups in the libyijinjing journal layer. All are behaviour-preserving on the wire: no frame bytes change, and the reader hands frames out in exactly the same order.

### writer constructors: 8 → 4

Four overloads existed only to accept a `lazy` flag the body discarded via `(void)lazy`. The journal-taking overload additionally ignored `low_latency` and `bus`, because the journal it receives has already fixed both. All twelve call sites passed `lazy=true`, so dropping the parameter preserves behaviour.

Worth recording for future retirements: these overloads were already marked `[[deprecated]]`, but the build carries `-Wno-deprecated-declarations`, so the marker never produced a single warning. Marking without a gate migrates nobody.

### reader merge order: say it once, in domain terms

`sort()` compared `gen_time` against an `INT64_MAX` sentinel that nothing ever lowered, so the branch was always taken — removed. The comparator was named `later` and used with `max_element` to select a *minimum*, into a variable named `min_journal_it`. It is now one `reads_before` definition (smallest `gen_time`, ties to higher `Priority`), with the heap parameterised on its transpose so `top()` reads as "next to read".

The invariant that makes the lazy `gen_time` comparison sound is now stated rather than implied: a journal parked in the heap must never advance, which is exactly why `sort()` pops the journal it hands to `current_` back into the recheck buffer.

### frame payload writes are now bounded

`reserve_frame()` is the only thing that bounds a payload write, but `frame::copy_data()` cannot see the reservation, and `close_frame()`'s assert fires *after* the memcpy and vanishes under `NDEBUG` — unchecked in exactly the build that matters. The checked entry point now lives on `frame_transaction`, which already holds the reservation; `write` / `write_as` / `write_at` route through it. `frame::copy_data()` remains the unchecked primitive and now says so.

`open_frame_lock_free()` / `close_frame_lock_free()` had no callers anywhere in the tree, and ADR-0063 already reserved `*_unserialized` for the internal single-owner primitive, so they are removed rather than re-documented.

### cloned_frame

`copy()` / `open()` overwrote `header_` without releasing the previous buffer — `replay_writer` calls `copy()` twice per replayed frame, so this leaked one allocation per frame — and wrote through `malloc`'s result with no null check. Both now go through `allocate()`, which raises on failure and frees first.

## Tests

Five reader-merge tests cover the k-way merge, priority tie-breaks, priority-over-`gen_time`, agreement between the scan and heap selection paths, and the contract that `join()` must not move a live cursor (previously unprotected, and precisely the kind of thing a merge-order refactor tidies away).

They were confirmed to fail against a deliberately inverted comparator before being accepted: three of the five turn red, so they are not vacuous.

## Verification

- `pnpm run configure` + `pnpm run compile` on framework/core: green, 0 errors
- `yijinjing_mmap_tests`: 25/25
- `yijinjing_content_hash_tests`: pass

Consumer migration was closed out by a full compile rather than by grep: three of the twelve call sites — a directly-constructed temporary, the pybind `py::init<>` template arguments, and a namespace-aliased `make_shared` — carry no greppable `writer(` text.

## Version impact

KFD-1: this registers `pykungfu-binding` as a welded surface. Python consumers bind to exported signatures at import time with no runtime negotiation, which is the register's own test, so its absence was a gap in the register rather than evidence the surface is not welded. Dropping `lazy` narrows `pykungfu.writer.__init__` from seven arguments to six — a break under the register, carried as pre-release: no in-tree Python constructs `writer`, and `stubs/pykungfu/` regenerates from the binding rather than being a hand-maintained contract. No line opens.

## Governance risk check

- [x] none of the above


dongkeren:
Closing and reopening from the development account. This PR was opened while the local gh active account had been left on kungfu-origin, which breaks the zone's two-account separation: kungfu-origin is the review/merge identity and cannot approve its own PR. Same branch, same commits, same verification — reopening as dongkeren so the review can happen properly.

PR #1022, opened 2026-07-17
fix(yijinjing): carry page lifecycle as policy instead of reading the environment

dongkeren:
Journal's constructor read `KF_KEEP_PAGE` and `KF_MAX_PRE_CREATE_SIZE` through
`getenv()`, punching a configuration side channel through the explicit policy
chain (`io_mapping_policy` -> `reader_policy` -> `journal_open_policy` ->
`page_open_policy`) that every other journal behaviour already flows through.

Page lifecycle now travels as a `page_lifecycle_policy` inside
`journal_open_policy`. Because `policy_` is const and is copied by the copy
constructor, two defects that followed from the side channel become impossible
by construction:

- the copy constructor carried `policy_` but not the env-seeded members, so a
  copied journal silently lost the lifecycle it had been created with;
- the destructor's `keep_page_ = false` was a dead store.

Parsing moves to a pure `parse_page_lifecycle()` that reports an unusable
setting as a value instead of swallowing it in a log, per the tier 3 policy in
`docs/development/cpp-error-handling.md`. `io_device` is now the single place
that consults the environment.

### Behaviour change

These are undocumented diagnostic knobs with no in-tree consumer (verified by
whole-repo grep: the only occurrences were the lines that read them). They now
apply to journals opened through `io_device` rather than to every journal ever
constructed. The presence-means-enabled contract for `KF_KEEP_PAGE` is
preserved, and both knobs are honoured for one migration window before they
retire.

### Verification

- `./shifu build:core` passed from a clean build tree.
- `./shifu test:mmap` passed 22/22, including new coverage for the pure parser
  and for lifecycle surviving a policy copy (the regression test for the
  copy-constructor defect).

### DIFF
--- a/framework/core/src/libkungfu/include/kungfu/runtime/io.h
+++ b/framework/core/src/libkungfu/include/kungfu/runtime/io.h
@@ -33,6 +33,11 @@ struct io_mapping_policy {
   [[nodiscard]] static constexpr io_mapping_policy from_legacy_lazy(bool lazy) noexcept {
     return lazy ? peer() : coordinator();
   }
+
+  [[nodiscard]] constexpr io_mapping_policy
+  with_lifecycle(yijinjing::journal::page_lifecycle_policy value) const noexcept {
+    return {reader.with_lifecycle(value), resource_manager_required, legacy_lazy_value};
+  }
 };
 
 FORWARD_DECLARE_CLASS_PTR(session)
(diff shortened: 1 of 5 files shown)


kungfu-origin (approved):

dongkeren:
Superseded by the current-dev linear replacement #1047, which preserved the page lifecycle policy implementation and mmap evidence, passed all required branch and merge-group checks, and merged as 8946c2b439256fa97cef7330f2c29bd218dbef3a.

PR #1041, opened 2026-07-17
fix(facts): queue concurrent material writers

### LINKED ISSUE #1030: facts: concurrent material writers fail with ownership_busy
## Summary

Two isolated Kungfu CLI processes writing different Fact Library subjects at the same time cannot both complete. One writer succeeds and the other exits with `RuntimeError: ownership_busy`.

This blocks the comparator's concurrent-agent tier. PostgreSQL and ClickHouse are exercised with the same requirement: two isolated clients start concurrently and both writes must be admitted without an adapter-side retry that would serialize the workload.

## Frozen input

- Kungfu source: `afce4713bec423a1760c7e1174e2400bc2f032ad`
- CLI version: `4.0.0-alpha.0`
- Package: `kungfu-episodes-cli-linux-x64.tar.gz`
- Package SHA-256: `3970615ddb708d91ab042548951b243f62fc725006fcd362cd71aecd62665798`
- build-images baseline: `v1.2.4-alpha.26` / `e3051187f2e1093d1e2b99d5cb8fe746c3a4351c`
- Runtime: Linux amd64, Docker, one shared Fact Library home

This is dev-main prequalification evidence, not an alpha release claim.

## Reproduction

1. Create Fact Surface `qualification-job-facts` v1 with sources `agent-a` and `agent-b`.
2. Admit an initial material.
3. From two independent processes, start these commands concurrently against the same `-H` directory, using different sources, subjects, and observation IDs:

```bash
kungfu -H /var/lib/kungfu/primary facts material put \
  --type qualification-job-facts --type-version v1 \
  --source agent-a --subject job-agent-a \
  --payload-file - --observation-id obs-agent-a --action assert

kungfu -H /var/lib/kungfu/primary facts material put \
  --type qualification-job-facts --type-version v1 \
  --source agent-b --subject job-agent-b \
  --payload-file - --observation-id obs-agent-b --action assert
```

## Actual result

One command returns a successful `kungfu.facts.material-write/v1` receipt. The other exits 1 with:

```text
RuntimeError: ownership_busy: another process owns /var/lib/kungfu/primary/runtime/ownership/writers/05014d64.00000000.lock
```

The result reproduced for all three comparator jobs (J1/J2/J3). The other six tiers—normal, SIGKILL recovery, whole-root export/import restore, historical correction/head query, non-overlapping v1/v2 schema evolution, and fresh-agent takeover—passed in the same package/runtime.

## Expected result

Both isolated writers complete and both admitted materials are visible in the catalog, or Kungfu provides a product-level blocking/queueing interface with a bounded contract that preserves true concurrent client admission. Adapter-side retry/serialization should not be required to hide ownership contention.

dongkeren:
## Summary

- acquire the Fact admission writer before reading current state, so concurrent clients are serialized at the Core commit boundary instead of racing through a read/write gap
- expose a bounded 5-second `bounded-core-wait/v1` writer-admission contract while preserving the yijinjing single-physical-writer invariant
- make concurrent CLI home-directory initialization idempotent
- add a real 8-process public CLI test proving all distinct Fact subjects are admitted and visible without adapter-side retry

Fixes #1030

## Local validation

- C++23 macOS build: `cmake --build framework/core/build --target kungfu pykungfu --parallel 20`
- `pytest framework/core/tests/python/test_storage_cli.py -k fact_cli_queues_real_multi_process_material_writers -vv`
- `./shifu check`

### DIFF
--- a/framework/core/src/libkungfu/src/runtime/facts/fact_admission.cpp
+++ b/framework/core/src/libkungfu/src/runtime/facts/fact_admission.cpp
@@ -7,17 +7,21 @@
 #include <kungfu/view/action_envelope.h>
 #include <kungfu/view/schema.h>
 #include <kungfu/yijinjing/journal/assemble.h>
+#include <kungfu/yijinjing/ownership.h>
 #include <kungfu/yijinjing/storage/content_hash.h>
 #include <kungfu/yijinjing/storage/episode_manifest.h>
 #include <kungfu/yijinjing/time.h>
 
 #include <algorithm>
 #include <charconv>
+#include <chrono>
 #include <limits>
 #include <map>
+#include <memory>
 #include <set>
 #include <stdexcept>
 #include <string>
+#include <thread>
 #include <utility>
 #include <vector>
 
@@ -31,6 +35,7 @@ namespace {
 constexpr uint32_t FACT_EVENT_SCHEMA_VERSION = 1;
 constexpr const char *FACT_NAMESPACE = "facts";
 constexpr const char *FACT_NAME = "admission";
+constexpr auto FACT_WRITER_WAIT_TIMEOUT = std::chrono::milliseconds(5000);
 
 struct schema_contract {
   view::schema_handle handle;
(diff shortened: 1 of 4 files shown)


kungfu-origin (approved):
Reviewed the bounded Core admission queue, CLI initialization race fix, and real multi-process qualification. The physical single-writer invariant remains intact and the client contract is explicit.

PR #1047, opened 2026-07-17
fix(yijinjing): carry page lifecycle as policy

dongkeren:
## Summary

- make page lifecycle an explicit runtime/journal policy rather than reading environment variables inside journal code
- keep the existing storage paths and defaults while making lifecycle selection testable
- supersede #1022 with a `fix/*` linear branch because the old `feature/*` name falsely required a feature-delivery ADR declaration

Supersedes #1022.

## Mac verification

- the original change passed a full `./shifu build:core` and 22/22 mmap tests on Mac
- the replacement rebased cleanly over #1024; the runtime IO source blobs remain byte-identical, while journal files contain the expected combined current-dev changes
- current native reconfiguration was attempted but correctly refused before compilation because this host has Ninja 1.10.2 while the current repository contract requires >=1.11.1
- `git diff --check` — passed; merge-queue affected-native is the current-base integration proof

### DIFF
--- a/framework/core/src/libkungfu/include/kungfu/runtime/io.h
+++ b/framework/core/src/libkungfu/include/kungfu/runtime/io.h
@@ -33,6 +33,11 @@ struct io_mapping_policy {
   [[nodiscard]] static constexpr io_mapping_policy from_legacy_lazy(bool lazy) noexcept {
     return lazy ? peer() : coordinator();
   }
+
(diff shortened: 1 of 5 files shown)


kungfu-origin (approved):
Reviewed the rebased page lifecycle policy change and preserved mmap qualification evidence. Current required checks are green apart from the still-running affected-native integration check.

PR #1076, opened 2026-07-18
fix(core): enforce declared event route closure

dongkeren:
## Summary

- Require every dynamic event route to declare an admitted extension point before subscription.
- Inventory every event subscription surface and fail closed on source drift.
- Cover all extension classes with runtime, static, and negative-fixture checks.

## Validation

- ./shifu build:core
- core CTest: 21/21 passed
- ./shifu docs:check:readonly
- ./shifu adr:audit
- route topology contract and self-test
- Project Cut exact-root closeout gate


dongkeren:
Refreshing the pull_request event after adding the required ADR delivery manifest.

kungfu-origin (approved):

PR #1181, opened 2026-07-21
feat(xinfa): add hash-pinned wasm engine

dongkeren:
## Summary

- compile the shared pure Xinfa core to a checked-in, hash-pinned wasm32 engine
- let Shifu use a zero-dependency Node host without runtime Rust compilation, with source freshness checks and native fallback
- stage and verify the same wasm as a release artifact while keeping production receipt minting native-only
- remap host build paths and reject private paths or addresses during qualification
- bind delivery to ADR-0128 and successor Project Cut 0791a08f

## Verification

- `./shifu xinfa:check`
- `./shifu xinfa:wasm:check`
- `node --test product/scripts/dist.test.mjs scripts/check-shifu-entry-contract.test.mjs xinfa/tooling/check-boundary.test.mjs`
- staged repository gate and deterministic docs gate
- exact-root Project Cut independent review: fit


kungfu-origin (approved):

dongkeren:
Superseded by #1183. The final branch preserves the same implementation while publishing each Project Cut predecessor in its own linear commit and adding rustfmt to the pinned minimal Rust toolchain, resolving Source Acceptance and all three platform Xinfa checks without force-pushing history.

PR #1183, opened 2026-07-21
feat(xinfa): add hash-pinned wasm engine

dongkeren:
## Summary

- compile the shared pure Xinfa core to a checked-in, hash-pinned wasm32 engine
- let Shifu use a zero-dependency Node host without runtime Rust compilation, with source freshness checks and native fallback
- stage and verify the same wasm as a release artifact while keeping production receipt minting native-only
- remap host build paths and reject private paths or addresses during qualification
- pin Rust 1.95 with wasm32 and rustfmt components for self-contained cross-platform checks
- publish a linear Project Cut chain ending at a70553d4

## Verification

- `./shifu xinfa:check`
- `./shifu xinfa:wasm:check`
- `./shifu check:source`
- `node --test product/scripts/dist.test.mjs scripts/check-shifu-entry-contract.test.mjs xinfa/tooling/check-boundary.test.mjs`
- exact-root Project Cut independent review: fit
- Atlas closeout gate: pass


dongkeren:
Superseded by #1185, which regenerates a single Project Cut on the current dev/v4/v4.0 base and passes the scoped composition gate.

PR #1259, opened 2026-07-22
fix(core): restore witnessed documentation material

dongkeren:
Restore the product Documentation Atlas from an immutable selector-pinned Git material commit when Buildchain uses a witness-only or shallow checkout. Verify every restored artifact against its tracked size and SHA-256 witness, add shallow-clone and tamper coverage, and refresh affected KFD evidence.\n\nADR impact: bugfix only; no architecture decision status change.


kungfu-origin (approved):

dongkeren:
Superseded: this branch was accidentally based on a divergent local dev ref. Reissuing the same reviewed fix from exact origin/dev/v4/v4.0 without rewriting published history.

PR #1270, opened 2026-07-22
fix(shifu): dispatch Windows lifecycle after resolution

dongkeren:
## Summary

Restore direct dispatch at every resolved Windows launcher path so ordinary lifecycle tasks reach pnpm after a cold, warm, explicit, or release-pinned resolution.

## Diagnosis

Alpha promotion Build run 29959438156 compiled the launcher and returned 0 without entering `dist`; the following `verify` found only the qualification summary instead of the required product artifacts. The `:runresolved` repair added after the last successful three-platform Build was the behavioral delta. Its tests exercised `--version` and the build-free `cache` shortcut, not an ordinary command.

## Changes

- dispatch the resolved executable directly with the original `%*` vector
- use delayed expansion only for binary paths first assigned inside parsed blocks
- add a Windows live probe that requires `doctor` output across cold, warm, explicit, and cache-applied paths
- update the static contract to reject reintroduction of the no-op `:runresolved` path

## Verification

- targeted entry tests: 3 passed
- lifecycle tests: 10 passed, 2 Windows-only skipped on macOS
- full staged repository gate, docs gate, Biome formatting, and DCO: passed

### DIFF
--- a/scripts/check-shifu-entry-contract.test.mjs
+++ b/scripts/check-shifu-entry-contract.test.mjs
@@ -164,32 +164,19 @@ test('cold source acquisition copies the binary from its isolated Cargo target',
   assert.doesNotMatch(windowsAcquire, /crates\\target\\release/);
 });
 
-test('Windows freshly resolved launchers dispatch outside parsed build blocks', () => {
+test('Windows launchers dispatch the original arguments at every resolution path', () => {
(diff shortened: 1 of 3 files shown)


kungfu-origin (approved):
Reviewed the Windows lifecycle regression proof, direct dispatch repair, and ordinary-command coverage.

PR #1272, opened 2026-07-22
fix(shifu): dispatch after Windows source build

dongkeren:
## Summary

Separate Windows source-launcher resolution from task dispatch so a successful cold build cannot return 0 before executing the original lifecycle command.

## Diagnosis

Alpha Product Build runs 29959438156 and 29961313640 both showed the same runtime signature: the launcher compiled successfully, no `doctor`/install/dist output followed, every later stage rebuilt the launcher, and `verify` found only one of eight required artifacts. The cache publication and task invocation were nested inside the post-build command block, so that block remained an execution authority and could silently skip the task.

## Changes

- record the fresh target executable inside the build block and dispatch only after the block closes
- treat publishing the warm cache slot as an optimization, not a prerequisite for running the task
- fall back to the just-built target executable when cache publication is unavailable
- ratchet the static contract around top-level post-build dispatch; retain the Windows live `doctor` probe

## Verification

- targeted source-launcher contract tests: 4 passed
- lifecycle tests: 10 passed, 2 Windows-only skipped on macOS
- full staged repository gate, docs gate, Biome formatting, and DCO: passed

### DIFF
--- a/scripts/check-shifu-entry-contract.test.mjs
+++ b/scripts/check-shifu-entry-contract.test.mjs
@@ -139,7 +139,9 @@ test('source-fresh launchers pin nested Shifu entry to the resolved binary', () 
     posix.match(/SHIFU_BIN="\$kungfu_dev_bin" exec "\$kungfu_dev_bin"/g)
       ?.length >= 2,
   );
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Reviewed the runtime evidence, top-level post-build dispatch, cache-publication fallback, and regression contracts.

PR #1274, opened 2026-07-22
chore(release): reconcile alpha promotion with dev

dongkeren:
## Summary

Promote the qualified v4 development channel to alpha and settle all 134 changed accepted ADR projections.

The reconciliation commit preserves both channel ancestries while its tree exactly equals current dev. It includes libnode `22.22.3-kf.4`, KFD aggregate provenance, and credential-free shallow restoration of the public Documentation Atlas material.

## Verification

- exact two-parent topology and tree equality with current dev
- ADR alpha settlement gate: 134 projections, zero findings
- repair PRs #1268, #1270, and #1272 passed merge-group qualification
- release publication starts only after this PR is merged

PR #1277, opened 2026-07-22
fix(shifu): inline Windows environment loading

dongkeren:
## Summary

Inline the two Windows build-local.env reads and remove the trailing `:loadenv` batch subroutine. Recursive Node shell entry could make `cmd.exe` lose that label and leak `cannot find the batch label specified - loadenv` into strict child protocols.

## Verification

- entry-contract and lifecycle tests: 26 passed, 2 Windows-only skipped locally
- full staged gate passed
- regression contract requires inline parsing and forbids `call :loadenv` / `:loadenv`
- failure reproduced in alpha promotion PR #1274, Windows shifu CI job 89066975673

ADR-neutral bugfix; no ADR status transition.

### DIFF
--- a/scripts/check-shifu-entry-contract.test.mjs
+++ b/scripts/check-shifu-entry-contract.test.mjs
@@ -132,6 +132,16 @@ test('Windows source-fresh launcher retries one transient build failure', () => 
   assert.match(sourceBuild, /if "!_KFC_BUILD_ERROR!"=="0" \(/);
 });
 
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Reviewed against the Windows stderr failure and label-free regression contract.

dongkeren:
Refreshing the pull_request event after adding the required ADR-neutral manifest; no source changes.

PR #1283, opened 2026-07-22
fix(runtime): bind installed contract search root

dongkeren:
## Summary

- pass the assembled product root into the native Action Runtime as its default `search_base`
- preserve caller-supplied search roots
- cover both installed-product injection and explicit override behavior

## Diagnosis

Alpha Product Build run 29965568435 successfully assembled and packaged Linux, including the contract registry under `runtime/config`, then failed the installed CLI capability smoke. `work_profile.capabilities()` crosses into the C++ Action Runtime with an empty runtime directory and no `search_base`; native registry discovery therefore searched only the qualification temp cwd. The Python host seam already knows the installed product root, so this patch carries that fact across the Python/native boundary without relying on CI-only environment variables.

## Verification

- `PYTHONPATH=src/python uv run --frozen pytest tests/python/test_contract.py -q` — 3 passed
- Ruff format and lint — passed
- full staged repository gate — passed

### DIFF
--- a/framework/core/src/python/kungfu/storage/service.py
+++ b/framework/core/src/python/kungfu/storage/service.py
@@ -988,8 +988,16 @@ def action_runtime(
     array, so the wrapper preserves non-object results instead of forcing dict().
     """
 
+    operation_request = {"action": action, **(request or {})}
+    if "search_base" not in operation_request:
+        from kungfu import host
+
+        product_root = host.product_root()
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Independent review: root cause matches the installed-product failure, the host seam is carried across the native boundary, and explicit search roots remain authoritative.

PR #1288, opened 2026-07-23
fix(shifu): preserve Windows pnpm task dispatch

dongkeren:
## Summary

Product Build run 29968856850 proved the Windows launcher could build successfully while ordinary lifecycle arguments never reached pnpm: install/build/verify emitted only launcher-build messages, and verify produced one artifact instead of the required eight.

Enter the repository batch shim through an explicit `call` command boundary under `cmd.exe /d /s /c`. This preserves the full ordinary-task argument vector after the batch file returns through nested cache-applied lifecycle execution.

The regression suite now executes a real Windows `pnpm exec node` command and requires a filesystem side effect. Previous coverage only exercised launcher-owned `--version` and `doctor` paths and therefore could not detect a dropped pnpm task.

## Verification

- `node --test scripts/run-shifu-lifecycle.test.mjs` — 11 pass, 3 Windows-only skips locally
- `node --test scripts/check-shifu-entry-contract.test.mjs` — 16/16 pass

### DIFF
--- a/scripts/run-shifu-lifecycle.mjs
+++ b/scripts/run-shifu-lifecycle.mjs
@@ -49,7 +49,11 @@ export function windowsCmdArgs(shim, args) {
       'Windows Shifu lifecycle arguments contain unsafe cmd syntax',
     );
   const quote = (value) => `"${String(value).replaceAll('"', '""')}"`;
-  const payload = [shim, ...args].map(quote).join(' ');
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Approved: the failure is grounded in Product Build logs, and the new Windows regression proves an ordinary pnpm command reaches a concrete side effect.

PR #1292, opened 2026-07-23
fix(shifu): preserve fallback lifecycle argv

dongkeren:
## Summary

Product Build run 29970417379 proved the previous Windows regression test was a false positive: the test process inherited `SHIFU_BIN` from its parent Shifu invocation and bypassed the cmd fallback, while the real cache-active Product Build still ran no ordinary pnpm tasks and produced one artifact.

This patch:

- passes `cmd.exe /d /s /c call`, the batch shim, and every task argument as discrete spawn arguments so Node owns Windows quoting;
- removes `windowsVerbatimArguments`, eliminating the hand-built single command payload that dropped the ordinary task vector;
- makes the Windows regression strip `SHIFU_BIN`, force `SHIFU_CACHE_ACTIVE=1`, use an isolated launcher cache, and require a real `pnpm exec node` filesystem side effect.

## Verification

- `node --test scripts/run-shifu-lifecycle.test.mjs scripts/check-shifu-entry-contract.test.mjs` — 27 pass, 3 Windows-only skips locally

### DIFF
--- a/scripts/run-shifu-lifecycle.mjs
+++ b/scripts/run-shifu-lifecycle.mjs
@@ -48,13 +48,10 @@ export function windowsCmdArgs(shim, args) {
     throw new Error(
       'Windows Shifu lifecycle arguments contain unsafe cmd syntax',
     );
-  const quote = (value) => `"${String(value).replaceAll('"', '""')}"`;
-  // `call` gives cmd.exe an explicit batch-file invocation boundary. Without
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Approved: the regression now forces the exact cache-active, no-SHIFU_BIN fallback path that failed in Product Build, and the spawn argv is delegated to Node quoting.

PR #1294, opened 2026-07-23
release: promote dev/v4/v4.0 to alpha

dongkeren:
## Summary

Promote the qualified v4 development channel to alpha and settle all 135 changed accepted ADR projections.

The reconciliation commit preserves both channel ancestries while its tree exactly equals current dev. It includes libnode `22.22.3-kf.4`, KFD aggregate provenance, and credential-free shallow restoration of the public Documentation Atlas material.

## Verification

- exact two-parent topology and tree equality with current dev
- ADR alpha settlement gate: 135 projections, zero findings
- repair PRs #1268, #1270, #1272, #1277, #1279, #1283, #1286, #1288, and #1292 passed merge-group qualification
- release publication starts only after this PR is merged

PR #1295, opened 2026-07-23
fix(shifu): quote Windows lifecycle payload once

dongkeren:
## Summary

The corrected cache-active fallback regression in alpha PR #1294 proved that passing discrete arguments after `cmd.exe /c` exits 255. The earlier single-payload form failed because it combined hand-written outer quotes with `windowsVerbatimArguments`.

This patch uses the Windows-safe middle ground:

- construct one `call "shifu.cmd" "task" ...` payload for `/c`;
- pass that payload as one normal spawn argument with no hand-written outer quote pair;
- let Node escape the single CreateProcess argument;
- retain the corrected regression that removes inherited `SHIFU_BIN`, forces `SHIFU_CACHE_ACTIVE=1`, isolates the cache, and requires a real pnpm filesystem side effect;
- expose child output during the regression so a future failure is diagnostic.

## Verification

- `node --test scripts/run-shifu-lifecycle.test.mjs scripts/check-shifu-entry-contract.test.mjs` — 27 pass, 3 Windows-only skips locally

### DIFF
--- a/scripts/run-shifu-lifecycle.mjs
+++ b/scripts/run-shifu-lifecycle.mjs
@@ -48,10 +48,13 @@ export function windowsCmdArgs(shim, args) {
     throw new Error(
       'Windows Shifu lifecycle arguments contain unsafe cmd syntax',
     );
-  // Keep the batch command and every task argument discrete so Node owns the
-  // Windows command-line quoting. A hand-built single /c payload survived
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Approved: the prior alpha run produced a concrete 255 counterexample; this patch keeps one /c payload, removes verbatim outer quoting, and retains the exact cache-active fallback regression.

PR #1299, opened 2026-07-23
fix(shifu): preserve raw Windows lifecycle payload

dongkeren:
## Summary

- pass the single `call ...` payload to `cmd.exe` with `windowsVerbatimArguments: true` so Node does not turn token quotes into literal backslash-quote bytes
- align Product Qualification coverage with the same raw payload contract
- retain the corrected Windows fallback regression that executes a real pnpm command and requires a filesystem side effect

## Verification

- `node --test scripts/run-shifu-lifecycle.test.mjs scripts/run-zero-burden-product-qualification.test.mjs` — 21 pass, 3 Windows-only skips locally
- Biome passed for both changed files
- full signed commit gate passed, including 92 documentation/promotion tests
- failing exact Windows evidence: run 29973082365, job 89099070087 (`\"...shifu.cmd\" is not recognized`)

## Governance risk check

- [ ] none of the above

This repair changes qualification invocation only; it does not publish artifacts.

### DIFF
--- a/scripts/run-shifu-lifecycle.mjs
+++ b/scripts/run-shifu-lifecycle.mjs
@@ -50,9 +50,9 @@ export function windowsCmdArgs(shim, args) {
     );
   const quote = (value) => `"${String(value).replaceAll('"', '""')}"`;
   // cmd.exe owns everything after /c as one command string. Keep that payload
-  // as one spawn argument, but do not add the outer quote pair ourselves:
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Independent repair review: the change addresses the observed Node quote escaping directly, keeps one cmd payload, and preserves a real fallback-side-effect regression.

PR #1323, opened 2026-07-23
fix(release): hydrate ADR promotion boundaries

dongkeren:
## Summary

Hydrate the exact alpha base and promotion head commits when ADR settlement runs inside Buildchain’s immutable depth-1 PR merge checkout, then retry the fail-closed changed-file computation.

The fetch accepts only two full 40-character commit OIDs from the trusted pull-request event and keeps the checkout shallow at depth 2.

## Verification

- shallow promotion fixture proves both boundaries are initially absent and exact hydration restores the ADR diff
- focused ADR release tests: 15/15 passed
- Biome check passed
- full staged commit hook passed, including documentation and invariant gates

### DIFF
--- a/scripts/adr-release-gate.mjs
+++ b/scripts/adr-release-gate.mjs
@@ -204,12 +204,40 @@ function stringList(value) {
 }
 
 /** @param {string} ref @param {string} head @param {string} root */
-function changedFilesBetween(ref, head, root) {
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Independent review: exact-OID shallow-boundary hydration remains fail-closed and is covered by a synthetic promotion checkout.

PR #1353, opened 2026-07-23
fix(ci): bypass Shifu bootstrap in Alpha preflight

dongkeren:
## Summary

- execute the pure Node receipt writer directly after `setup-node`
- pin Node 24 for aggregate/verify as well
- keep the platform probe credential-free and avoid Shifu toolchain bootstrap
- refresh the closed-world workflow authority projection

## Failure restored

After #1347 fixed the initial argv boundary, exact-source run https://github.com/kungfu-systems/kungfu/actions/runs/30012390934 exposed the next failure on Windows: Shifu attempted to bootstrap fnm and Git Bash tar could not extract the `C:` temporary path. macOS was then cancelled by fail-fast; Linux passed.

The receipt script itself has no package-manager dependency, so this patch enters it through the already-pinned Node 24 runtime on all hosted platforms.

## Verification

- `./shifu test:alpha-promotion-preflight` (10/10)
- direct platform receipt generation
- `./shifu gate:workflow-authority:refresh`
- `./shifu check:source`
- commit staged gate

### DIFF
--- a/.github/workflows/alpha-promotion-preflight.yml
+++ b/.github/workflows/alpha-promotion-preflight.yml
@@ -116,7 +116,8 @@ jobs:
           PREFLIGHT_PLATFORM: ${{ matrix.platform }}
         run: |
           mkdir -p .buildchain/alpha-promotion-preflight
-          ./shifu alpha:promotion:preflight write-platform \
+          # shifu-entry-contract: allow receipt generation to bypass the platform-specific Shifu bootstrap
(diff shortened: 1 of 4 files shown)


kungfu-origin (approved):

dongkeren:
Superseded by #1351, which applies the same hosted-platform receipt fix with the smaller probe-only change and has now merged. Local verification remains available on this PR; branch retained.

PR #1389, opened 2026-07-24
fix(core): stabilize generated binding annotations

dongkeren:
## Summary

- canonicalize host-dependent unresolved pybind11 stub annotations to `typing.Any`
- preserve function-body and default-value ellipses
- add the normalization case to the Core tooling suite

## Failure evidence

- alpha Build run `30058447310` built successfully but release verification failed because macOS stub generation changed tracked `runtime.pyi` from `typing.Any` to `...`
- the patch was checked against the exact dirty stub retained by that runner; normalization reproduces the committed canonical bytes

## Verification

- Core tooling tests: 19 passed
- `node node_modules/typescript/bin/tsc -p tsconfig.tools.json`
- `node --test framework/core/tests/gen-stubs.test.js`
- full `./shifu check:source` reached the final tooling type-check and the reported TS7006 was fixed and rechecked
- staged repository gate

### DIFF
--- a/framework/core/.gyp/gen-stubs.js
+++ b/framework/core/.gyp/gen-stubs.js
@@ -19,6 +19,23 @@ const path = require('node:path');
 const glob = require('glob');
 const { shell } = require('../lib');
 
+/**
+ * @param {string} text
+ * @returns {string}
+ */
+function normalizeStubText(text) {
+  return (
+    text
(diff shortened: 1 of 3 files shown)


kungfu-origin (approved):
Independent review: the patch preserves clean-source verification by canonicalizing only unresolved type annotations, with focused regression coverage.

PR #1392, opened 2026-07-24
fix(shifu): keep Windows quality dispatch reachable

dongkeren:
Restore the Windows `xinfa:quality` dispatch used by the required merge-queue workspace check. The implementation block is mechanically moved next to the route table, and the entry contract now preserves that placement.

Validation:
- `./shifu xinfa:quality --check`
- Shifu entry/lifecycle contract tests
- `./shifu check:source`
- pre-commit gates


kungfu-origin (approved):
Reviewed as a narrow Windows launcher queue-unblocking fix with a targeted regression assertion.

PR #1395, opened 2026-07-24
fix(shifu): bypass adjacent quality target

dongkeren:
Restore ordinary Windows command dispatch after the `xinfa:quality` target was moved next to the route table. Without the explicit bypass, `shifu.cmd gate run ...` falls through into the quality argument parser.

This adds one `goto projectcut` and a regression assertion that preserves both routes.

Validation:
- Shifu entry/lifecycle contract tests
- `./shifu check:source`
- pre-commit gates
- failure reproduced from Windows merge-group run `30062072460`


kungfu-origin (approved):
Reviewed against Windows merge-group failure 30062072460; the explicit bypass restores the prior fallthrough route.

PR #1495, opened 2026-07-25
fix(gui): keep global Work observer app-scoped

dongkeren:
## Summary

Keep the installed global Work observer truly app-scoped after live dogfood exposed renderer subscription churn and orphaned recovery workers.

## Changes

- retain the single main-process observer when the renderer view unsubscribes or remounts
- keep the latest shell notification target in a ref so renderer re-renders do not restart observation
- handle `SIGTERM` in the observer CLI and let an active ProcessPool close before exit
- prove unsubscribe/resubscribe reuses one child and app disposal performs the one shutdown

## Verification

- `./shifu test:work-authority`
- 39 workspace federation Python tests
- 3 main-process and renderer observer tests
- GUI production build
- KFD buildchain check
- live SIGTERM during an active recovery ProcessPool: exit 0, orphaned children 0

## Governance risk check

- [x] none of the above

## Checklist

### DIFF
--- a/extensions/work-dashboard/src/view/index.tsx
+++ b/extensions/work-dashboard/src/view/index.tsx
@@ -137,6 +137,8 @@ function GlobalWorkView({
   const [status, setStatus] = React.useState('connecting live global Work…');
   const [error, setError] = React.useState('');
   const lastNotification = React.useRef({ key: '', at: 0 });
+  const shellRef = React.useRef(shell);
+  shellRef.current = shell;
   const ipc = React.useMemo(
     () =>
       (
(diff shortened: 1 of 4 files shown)


kungfu-origin (approved):
Independent review: exact head 74d5ea0ea6c971b53585b71e734187a695673ad6 closes renderer subscription churn and SIGTERM ProcessPool cleanup with targeted and live evidence.

dongkeren:
Closing and reopening from a fix/* head: this is an ADR-neutral live-dogfood bugfix, and the governance gate correctly rejected the inherited feature/* branch shape.

PR #1498, opened 2026-07-25
fix(workspace): keep repeated catalog observations stable

dongkeren:
## Summary

Keep the app-scoped global Work observer incremental across separate semantic writes by making repeated observation of an unchanged Workspace locator write-free.

## Changes

- preserve the existing Catalog entry, epoch, bytes, and cut when a qualified locator is semantically unchanged
- continue updating the Catalog when identity, locator, availability, or candidate replacement actually changes
- avoid persisting read-only Catalog projection fields during real updates
- prove repeated observation does not reorder entries or mutate Catalog bytes

## Verification

- ./shifu check:source
- ./shifu test:workspace-continuation with the exact 700b1214a5 native artifact: Node 8/8, Python 76/76
- staged commit gate

## Governance risk check

- [x] none of the above

## Checklist

### DIFF
--- a/framework/core/src/python/kungfu/workspace.py
+++ b/framework/core/src/python/kungfu/workspace.py
@@ -613,16 +613,53 @@ def observe_workspace_locator(
     if existing is not None:
         entry["lifecycle"] = _catalog_lifecycle(existing)
         entry["provenance"] = _catalog_provenance(existing)
-    entries = [
-        row
-        for row in catalog["entries"]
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Independent review of exact head 9a5c04c8a7b0377bbd415002f6d9a16e113ea960: idempotent no-write path preserves Catalog bytes and cut while real locator changes retain the existing update path; regression tests and source acceptance are adequate.

PR #1499, opened 2026-07-25
fix(agent-hub): isolate installed adapter variant

dongkeren:
## Summary

Keep the embedded Node-only variant marker inside the KFD execution process so the Agent Hub qualification can re-enter the exact installed Kungfu product as its adapter.

## Root cause

Auditable-demo build run 30175240748 reached the installed `kungfu agent hub qualify` smoke, then the KFD runner launched the packaged Kungfu front door while `KUNGFU_AS_VARIANT=node` was still inherited. The Rust trunk therefore treated the first public argument `agent` as a Node module path and failed with `Cannot find module .../agent`.

## Changes

- clear `KUNGFU_AS_VARIANT` at the fresh installed-product subprocess boundary
- preserve the internal KFD step envelope and all public Agent Hub arguments
- add a regression test that starts from a deliberately polluted parent environment

## Verification

- `PYTHONPATH=framework/core/src/python uv run --project framework/core --frozen pytest -q framework/core/tests/python/test_agent_hub_profile.py` (8 passed)
- `./shifu node --test product/scripts/dist.test.mjs` (23 passed)
- full pre-commit repository gate

## Governance risk check

- [ ] none of the above

This changes only the environment inherited by a build-time installed-archive qualification subprocess. It adds no publication, deployment, package write, secret, or identity-token authority.

## Checklist

- [x] Exact failed build evidence cited

### DIFF
--- a/framework/core/src/python/kungfu/agent/agent_hub_qualification.py
+++ b/framework/core/src/python/kungfu/agent/agent_hub_qualification.py
@@ -124,6 +124,12 @@ def run_kfd_step(entry: Path, *commands: str) -> None:
 
 def _run_kfd(executable: Path, entry: Path, *commands: str) -> None:
     env = os.environ.copy()
+    # The embedded Node host may mark its own process as the node variant.  That
+    # marker must not cross the fresh installed-product boundary: the KFD runner
+    # launches this executable again as the Agent Hub adapter, and the trunk
+    # would otherwise interpret the first public argument ("agent") as a Node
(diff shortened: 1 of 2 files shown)


dongkeren:
Superseded by the linear, current-base fix in #1500. The newer baseline intentionally enters the exact KFD script through the Python-free Node variant, so the final repair scopes that variant to the exact entry instead of clearing it globally.

PR #1501, opened 2026-07-25
fix(gui): keep installed bundle immutable

dongkeren:
## Summary

Keep the installed Developer ID-signed app bundle immutable while its global Work observer runs.

## Changes

- force the observer child process to disable Python bytecode writes
- preserve all inherited observer environment values
- prove a conflicting caller value is overridden and unrelated values survive

## Verification

- framework/gui/node_modules/.bin/tsx --test framework/gui/src/main/global-work-observer-host.test.ts (2/2)
- ./shifu check:source
- staged commit gate

## Governance risk check

- [x] none of the above

## Checklist


kungfu-origin (approved):
Independent review: the change is narrowly scoped, preserves inherited environment values, overrides bytecode generation explicitly, and includes a direct regression assertion. Fit for merge.

PR #1502, opened 2026-07-25
fix(gui): protect packaged Python resources

dongkeren:
## Summary

Set Python bytecode suppression at the packaged desktop entry before any child process can touch signed bundle resources.

## Changes

- protect the packaged desktop environment before main-process startup work
- preserve development behavior while retaining the observer-level defense
- cover packaged override, existing environment preservation, and development behavior

## Verification

- targeted desktop test: 3/3 passed
- `./shifu check:source` passed
- startup-level experiment kept strict deep codesign valid after the live observer started


kungfu-origin (approved):
Independent review: the packaged-entry environment guard is minimal, preserves development behavior, and closes the signed-resource mutation path without changing Work authority or observer semantics.

PR #1520, opened 2026-07-26
fix(sdk): use canonical engage subprocess path

kungfu-origin:
## Summary

- route SDK Python-AOT PDM and Nuitka subprocesses through the sole canonical `kungfu dev engage` path
- route Nuitka data-composer and SCons re-entry through the same canonical path
- add a regression that rejects the removed top-level `kungfu engage` path in internal subprocess owners

This fixes the real product.verify-full failure exposed after the Dev Patrol checkout repair and is a follow-up to #1061.

## Testing

- `./shifu --filter @kungfu-tech/sdk build` (32/32)
- targeted CLI surface regression (1/1)
- `./shifu check:source` (646 Node tests, 40 Python tests, 116 upgrade tests, 69 desktop tests; Ruff, Biome, mypy passed)

### DIFF
--- a/developer/sdk/src/sdk.js
+++ b/developer/sdk/src/sdk.js
@@ -2733,7 +2733,7 @@ function pythonAotBuild(manifest) {
   // legacy pdm-pep517 backend again for every extension build.
   runOrFail(
     py,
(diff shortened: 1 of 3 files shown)


kungfu-origin:
Closing and recreating from the same reviewed head under the designated author identity so the independent reviewer can approve it.

PR #1533, opened 2026-07-26
fix(verify): restore dev patrol contract coverage

dongkeren:
## Summary
- realign Dev Patrol fixtures with the current schema, provider set, read-only global Work view, and single Work mutation authority
- preserve report, receipt, and remote-sync coverage through externally authoritative Work references instead of restoring the retired WorkStore
- return the Fact kernel contract-specific relation endpoint failure and refresh its rooted generated evidence

## Verification
- all nine previously failing fixtures pass locally
- storage Node binding: 19/19 tests pass
- staged pre-commit gate passes, including C++ format, Python lint/format, KFD evidence, and documentation checks
- `./shifu build:core` and `./shifu freeze` pass

Follow-up for #1516. The standing Dev Verify Patrol will close the issue only after the merged mainline is green on all supported platforms.


kungfu-origin (dismissed):
Reviewed the current authority cutover and reproduced the failing fixtures. The patch preserves the single Work mutation authority, keeps report and remote facts source-scoped, fixes the contract-specific Fact failure code, and refreshes all rooted projections.

kungfu-origin (approved):
Reviewed the lint-portable fixture import follow-up and the complete Patrol contract closure.

PR #1777, opened 2026-07-28
fix(ci): install CodeBuild platform optionals

dongkeren:
## Summary

- let real AWS CodeBuild Linux qualification install the matching platform optional packages
- keep `--no-optional` unchanged for every other runner and lifecycle
- add a focused contract test for the clean-runner libnode dependency path

## Verification

- `node --test scripts/buildchain-install.test.mjs` (8/8)
- `./shifu check:source` (721 tests: 711 passed, 10 skipped)
- `git diff --check origin/dev/v4/v4.0...HEAD`

## Delivery boundary

Trusted exact-source qualification only. No release credentials or static cloud credentials are admitted.


kungfu-origin (approved):
Reviewed for the bounded AWS CodeBuild Linux qualification scope: matching platform optionals are admitted only on the real CodeBuild Linux path, while existing runner behavior and credential boundaries remain unchanged.

kungfu-origin (approved):

PR #1969, opened 2026-07-30
fix(runtime): detach Windows resident supervisor

dongkeren:
## Summary

- detach the Windows resident supervisor from kill-on-close parent jobs while retaining a controlled fallback for hosts that forbid breakaway
- preserve the existing detached process group semantics on every Windows host
- include bounded supervisor and coordinator log tails when product runtime qualification times out

## Evidence

- `PYTHONPATH=framework/core/src/python uv run --project framework/core --frozen pytest -q framework/core/tests/python/test_runtime_service.py` (37 passed)
- `./shifu exec node --test framework/core/tests/qualification/runtime-activation/run.test.mjs` (16 passed)
- Ruff format and lint passed
- full staged pre-commit gate passed

This fixes the reproducible Windows product smoke failure seen after `runtime restart`: cold activation and warm reuse passed, then the replacement supervisor exited at the spawning command boundary.

### DIFF
--- a/framework/core/src/python/kungfu/runtime_service.py
+++ b/framework/core/src/python/kungfu/runtime_service.py
@@ -1536,10 +1536,28 @@ class ProcessRuntimeHost:
                 new_process_group = getattr(
                     subprocess, "CREATE_NEW_PROCESS_GROUP", 0x00000200
                 )
-                kwargs["creationflags"] = detached_process | new_process_group
+                breakaway_from_job = getattr(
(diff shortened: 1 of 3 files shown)


kungfu-origin (approved):
Verified the bounded Windows resident-process change, local regression coverage, exact-source preflight, and run 30556530211 evidence: product-runtime-smoke passed in 9.56s and runtime-activation qualified. The later layers.surfaces failure is independent of this change.

kungfu-origin (approved):

kungfu-origin (approved):

PR #1971, opened 2026-07-30
fix(runtime): isolate frozen supervisor process

dongkeren:
## Summary

- reset PyInstaller state only when a frozen Kungfu runtime launches its long-lived supervisor
- preserve ordinary Python, coordinator, worker, and non-frozen process behavior
- prove the frozen launch boundary passes `PYINSTALLER_RESET_ENVIRONMENT=1` without mutating the source environment

## Root cause

Alpha Build run `30539085997` reproduced the Windows Product runtime smoke failure on attempts 3 and 4 after 72/72 build-chain verification passed. The frozen `kungfu.exe` launched the supervisor as an independent process, but PyInstaller continued to treat it as a child worker of the expiring parent application instance; the supervisor then stopped by signal before the coordinator became available.

## Evidence

- `PYTHONPATH=framework/core/src/python ./shifu exec uv run --project framework/core --frozen pytest -q framework/core/tests/python/test_runtime_service.py` — 37 passed
- `./shifu check:source` — passed, including 798 passed / 10 conditional skips in the source-acceptance contract suite
- `TMPDIR=/private/var/folders/lh/63dh4_t12tl7smhvvcr9wwlr0000gn/T ./shifu check` — passed
- targeted macOS temp-path normalization fixture — passed

This PR intentionally leaves installed Product state untouched. The decisive qualification is the protected Windows Alpha Product runtime smoke after this exact head reaches the dev candidate pipeline.

### DIFF
--- a/framework/core/src/python/kungfu/runtime_service.py
+++ b/framework/core/src/python/kungfu/runtime_service.py
@@ -8,6 +8,7 @@ import os
 import platform
 import signal
 import subprocess
+import sys
 import threading
 import time
 from contextlib import contextmanager
@@ -558,6 +559,15 @@ def command_env(
                 }
             )
     return env
+
+
+def _independent_process_env(env: Mapping[str, str]) -> dict[str, str]:
+    """Detach a long-lived child from the current frozen application instance."""
+
+    child_env = dict(env)
+    if getattr(sys, "frozen", False):
+        child_env["PYINSTALLER_RESET_ENVIRONMENT"] = "1"
+    return child_env
 
 
 def coordinator_run_command(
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Independent review of exact head 9288fbc85332ce1867cacb9abd2caa334d19c96a: the change is bounded to the frozen supervisor launch boundary, preserves non-frozen and coordinator semantics, and adds direct plus launch-path regression coverage. Approved for protected dev qualification.

kungfu-origin (approved):
Independent review on exact head 121d5c3a57d2db5aeb19f1c95e9867a172111dab: bounded PyInstaller child-process isolation; source and full repository checks verified.

kungfu-origin (approved):

PR #2111, opened 2026-08-01
fix(work): resolve retained seal successors

dongkeren:
## Summary

Repair retained Assignment seal resolution so exact typed endpoints and one strictly dominating successor remain readable without weakening fail-closed ambiguity.

## Changes

- Bind retained seals to a path-free Assignment semantic state.
- Collapse equivalent retained witnesses and select only a unique component-wise strict successor.
- Preserve selected, equivalent, and superseded state roots in Catalog evidence.
- Keep equal and incomparable semantic states unresolved.

## Verification

- ./shifu check:source
- focused Assignment orchestration and workspace federation tests
- KFD support matrix and Primitive profile
- exact retained Catalog source query

## Evolution Map impact

Evolution impact: extends

## Governance risk check

- [x] none of the above

## Checklist


dongkeren:
Closing as superseded by #2113. The merged successor carries the same retained-seal implementation and test delta across the four core files and the same 20-file delivery surface, was independently approved, and landed on dev/v4/v4.0 as cb566eb5e7aa273e0c532cd95b159ceb4ea3ae2d.

PR #2156, opened 2026-08-01
test(agent-session): make hosted portability fixtures neutral

dongkeren:
## Summary

- keep POSIX signal fixtures deterministic instead of inheriting the hosted runner platform
- let interaction-port tests opt into an explicit platform while defaulting to the reviewed POSIX contract
- validate the resolved node-pty path by path components on both Windows and POSIX

## Evidence

The exact Windows hosted Product qualification exposed three test-only portability failures after the production Windows PTY contract had already landed: two POSIX SIGINT fixtures inherited win32, and one valid node-pty path was compared with POSIX separators. Production signal semantics remain unchanged and fail closed for unsupported Windows signals.

## Verification

- ./shifu build:core: passed on protected base 77129be119d47ef467ec79e02164f778563ffe6b
- Agent Session suite with installed-version probes skipped: 122 passed, 0 failed
- focused three-file Biome check: passed
- pre-commit repository gate: passed
- full-repository lint reports four pre-existing errors outside this patch

## Evolution Map impact

Evolution impact: none

## Governance risk check

This test-only repair adds no credentials, signing authority, publishing authority, hosted-service access, or public release claim.

### DIFF
--- a/framework/agent-session/tests/capsule-host.test.mjs
+++ b/framework/agent-session/tests/capsule-host.test.mjs
@@ -38,7 +38,7 @@ class FakePtyProcess extends EventEmitter {
   }
 }
 
-function fixture(maxOutputBytes = 64, platform = process.platform) {
+function fixture(maxOutputBytes = 64, platform = 'linux') {
   const child = new FakePtyProcess();
   let now = 1000;
   const host = new AgentSessionCapsuleHost({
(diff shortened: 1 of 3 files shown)


dongkeren:
Superseded by PR #2153, which independently delivered the same hosted portability repair and merged through the protected queue as 0675b01877fbc8e3e865df258827e89141198ae1. Closing this duplicate without merging.

PR #2193, opened 2026-08-02

kungfu-origin:
## Summary

- route source-acceptance temporary state and nested Shifu diagnostics outside protected checkouts
- add adversarial cold read-only coverage for transitive package-manager, cache, diagnostic, and generated-fixture routes
- extract the platform dependency lifecycle from product/scripts/dist.mjs into one cohesive runtime-pin owner, retaining 342 lines of headroom

## Validation

- exact-checkout Core build bound to 5193abee554b09ecbf5df9dbbb86218165e29cae
- semantic amplification projection: issues=0
- local exact-root source acceptance and protected three-platform qualification are required before merge

### DIFF
--- a/framework/maintainability/readonly-source-routes.json
+++ b/framework/maintainability/readonly-source-routes.json
@@ -12,7 +12,19 @@
       "command": "./shifu check:source",
       "classification": "read-only-source",
       "implementation": "scripts/source-acceptance.mjs",
-      "transitiveWriters": [],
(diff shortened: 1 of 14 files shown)


dongkeren:
Closing as superseded by #2235. The latest replay is the active continuation, covers every path changed here, preserves the identical read-only pytest semantic patch, and adds the reviewed readonly-route fixes. The head branch is intentionally retained for recovery.

PR #2212, opened 2026-08-02

dongkeren:
Close protected source-acceptance write paths, retain an auditable byte-for-byte cold read-only fixture, and restore product assembly maintainability headroom without changing public product or architecture contracts.

Supersedes #2204 after the protected base advanced.


dongkeren:
Closing as superseded by #2235. The latest replay is the active continuation, covers every path changed here, preserves the identical read-only pytest semantic patch, and adds the reviewed readonly-route fixes. The head branch is intentionally retained for recovery.

PR #2217, opened 2026-08-02

dongkeren:
Close protected source-acceptance write paths, retain an auditable byte-for-byte cold read-only fixture, and restore product assembly maintainability headroom without changing public product or architecture contracts.

Supersedes #2212 after the protected base advanced.


dongkeren:
Superseded by #2224, which replays the identical two-commit patch onto the current `dev/v4/v4.0` base with an exact `git range-diff`. Retaining this PR and its successful checks as historical evidence.

PR #2235, opened 2026-08-02

dongkeren:


dongkeren:
Superseded by #2246, which replays the reviewed source change onto current dev/v4/v4.0 and preserves the updated exact-source and executable-closure qualification path. Historical CI evidence on this PR is retained; no force update was used.

PR #2246, opened 2026-08-02

dongkeren:


dongkeren:
Superseded by #2248, which replays the reviewed protected-source closure onto current dev/v4/v4.0 base a2967f22 with fresh exact-checkout source and Alpha evidence. Historical runs and branch are retained.

PR #2250, opened 2026-08-02
fix(tui): settle contextual project restore

dongkeren:
## Summary

Fix TUI startup restoration for a contextual Project so the restore is not cancelled by its own `loading` to `project-work` surface transition.

## Related issue

None.

## Changes

- Freeze the startup surface used to decide whether contextual Project restoration owns the initial request.
- Keep the live surface as the fail-closed commit gate for explicit user navigation.
- Add real Ink/React async regressions for the owned transition and explicit-navigation cancellation, including loading settlement.

## Verification

- `./shifu --filter @kungfu-tech/tui exec tsx --test src/control-plane.test.ts` (26/26)
- `./shifu build:cli`
- `./shifu test:kfx-profile-suite`
- `./shifu check:staged`

## Evolution Map impact

Evolution impact: none

## Governance risk check

- [x] none of the above

## Checklist

- [x] Documentation updated if behavior changed (not required; this restores intended behavior)

### DIFF
--- a/framework/tui/src/control-plane.test.ts
+++ b/framework/tui/src/control-plane.test.ts
@@ -24,7 +24,6 @@ import {
   reduceControlPlaneInput,
   resolveProductStartupSurface,
   shouldStartContextualProjectRestore,
-  useStartupProductSurface,
 } from './profile-shell.js';
 import {
   PROJECTS_QUICK_COMMANDS,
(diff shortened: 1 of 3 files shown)


kungfu-origin (dismissed):
Reviewed the startup-surface ownership change and both async Ink regressions. The live surface still revokes stale commits, while the current restore request now settles loading.

kungfu-origin (approved):
Re-reviewed the source-budget correction. The startup surface remains request-local, profile-shell no longer expands, and the exact complexity gate now passes.

PR #2414, opened 2026-08-03
fix(source): close protected checkout terminal paths

dongkeren:
## Summary

- route source-acceptance temporary state, diagnostics, caches, and nested task re-entry outside the checkout
- make the read-only source inventory transitive and fail closed on repository-local writers
- preserve product assembly ownership and durable line-budget headroom without a second authority
- retain the noninteractive demo transport and symlinked Python runtime materialization contract already present on current Dev

## Validation

- exact checkout `./shifu build:core`: passed
- cold locked discovery fixture: passed; before/after filesystem audit roots identical
- `product/scripts/dist.mjs`: below the 2200-line acceptance ceiling
- semantic amplification / abstraction integrity: zero blocking findings
- complete `./shifu check:source`: running on the exact candidate with the declared external read-only Python toolchain

## Delivery boundary

Candidate-head checks establish protected PR admissibility. Because Dev is moving, terminal completion remains fail-closed until the protected delivered tree itself passes post-merge source preflight, exact-root Linux/macOS/Windows qualification, and different-actor review.

### DIFF
--- a/framework/maintainability/readonly-source-routes.json
+++ b/framework/maintainability/readonly-source-routes.json
@@ -12,7 +12,19 @@
       "command": "./shifu check:source",
       "classification": "read-only-source",
       "implementation": "scripts/source-acceptance.mjs",
-      "transitiveWriters": [],
+      "runtimeImplementation": "scripts/readonly-source-toolchain.mjs",
+      "writerOwner": "source-acceptance-runtime",
+      "transitiveWriters": [
+        "disposable OS test directories only"
+      ],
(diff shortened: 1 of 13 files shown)


dongkeren:
Superseded append-forward by https://github.com/kungfu-systems/kungfu/pull/2416 after the managed pre-push gate correctly rejected rewriting the published replay35 branch. The exact rebased source is unchanged; replay35 and its checks remain historical evidence.

PR #2455, opened 2026-08-04
fix(qualification): serialize source suites before product rebuild

dongkeren:
## Summary

- serialize source/runtime qualification suites before product-distribution rebuilds the shared Core output
- preserve bounded parallelism inside each safe qualification phase
- prevent profile-action-admission from racing a concurrent cleanup of framework/core/build

## Related issue

Atlas/Kungfu Assignment 2026-08-01-kungfu-protected-source-terminal-closure.

## Changes

- split runtime activation execution into source suites, product producer, and product consumer phases
- retain declared result order and sibling-failure reporting
- add a scheduler contract test that proves every source suite settles before the product rebuild starts

## Verification

- reproduced the prior Linux x64 Full Build failure from run 30867735276: profile-action-admission lost pykungfu while product-distribution rebuilt the shared Core output
- ./shifu exec node --test framework/core/tests/qualification/runtime-activation/run.test.mjs (21/21 passed)
- ./shifu check (passed)
- authoritative Assignment worktree tree equals the validated worktree tree: 68e2674c3e0d8de59278b0722d97b26e54345b0c

## Evolution Map impact

Evolution impact: none

## Governance risk check

- [ ] credentials, tokens, secrets, or private logs
- [ ] provider APIs, CLIs, OpenTelemetry, billing, quota, or usage attribution
- [ ] official hosted or managed services
- [ ] official branding, package names, release identity, or domains
- [x] release evidence, provenance, package publishing, or deployment surfaces
- [ ] none of the above

Boundary: qualification ordering only. The final claim remains bound to a fresh protected merge and exact Linux/macOS/Windows Build receipts.

## Checklist

- [x] Commits are signed off (DCO)
- [x] Documentation is unchanged because the externally visible behavior and architecture contract are unchanged

PR #2460, opened 2026-08-04
fix(qualification): serialize source suites before distribution

dongkeren:
## Summary

Prevent runtime activation source suites from racing product distribution while both use the shared Core build layout.

## Root cause

The Linux x64 Formal Build ran profile-action-admission concurrently with product-distribution. Distribution rebuilt the shared Core layout after pykungfu had been assembled, so the profile suite transiently lost the exact current-checkout binding.

## Changes

- complete all source and runtime suites before product distribution mutates the shared layout
- run product distribution as an explicit barrier
- fan out product consumers only after distribution settles
- retain every existing suite and failure result

## Verification

- node --test framework/core/tests/qualification/runtime-activation/run.test.mjs (21/21)
- repository commit-time staged gate
- 73 Dev-gate tests, 17 invariant tests, and 138 documentation tests
- native Assignment run gate with exact Work Control and Dogfood Profile roots

Evolution impact: none

Governance risk: release qualification evidence is affected; verification keeps every required gate and makes their execution order deterministic.

- [x] Documentation is not required because the public behavior and contracts are unchanged.

### DIFF
--- a/framework/core/tests/qualification/runtime-activation/run.mjs
+++ b/framework/core/tests/qualification/runtime-activation/run.mjs
@@ -26,6 +26,7 @@ const PRODUCT_CONSUMER_SUITES = new Set([
   'product-verification',
   'product-catalog',
 ]);
+const PRODUCT_DISTRIBUTION_SUITE = 'product-distribution';
 
 export const SUITES = [
   {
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Reviewed the bounded DAG change and regression coverage. The source/profile suites now settle before the shared distribution build layout mutates, with no gate removed.

dongkeren:
Superseded by protected Dev commit 0c9154d3767389afe7b3be7eceb30042fc97d531 from merged PR #2455. That change implements the same three-stage source suites -> product distribution -> consumers barrier and carries the equivalent event-order regression test. The branch and run history are retained; no force update or deletion is performed.

PR #2582, opened 2026-08-06
fix(release): bind exact recovery Gate controller

dongkeren:
## Summary

- add an exact recovery Gate controller SHA to the bounded Alpha 1 dispatch
- pass that SHA to the refreshed Buildchain recovery train
- execute the read-only consumer Gate from the reviewed controller checkout while validating the immutable product subject separately
- bind subject and controller registries and refresh workflow authority evidence

## Safety

The existing candidate run `31051528142`, target SHA/tree, preflight, candidate artifacts, and payload bytes are unchanged. This adds no product build, install, repackaging, or manual release path.

## Validation

- targeted release/Gate tests: 44/44
- `./shifu check:staged`
- full commit hook gate
- explicit Gate aggregate probe qualified all four declared gates against the immutable target subject

## Runtime

`train/v3/v3.0/resume-candidate-run` resolves exactly to `66d360b1a91d48df468c930521782a1839934deb`.

### DIFF
--- a/.github/workflows/release-new-version.yml
+++ b/.github/workflows/release-new-version.yml
@@ -58,6 +58,11 @@ on:
         required: false
         default: ""
         type: string
+      resume-publication-gate-controller-sha:
+        description: "Exact Kungfu commit that owns the read-only recovery Gate controller"
+        required: false
+        default: ""
+        type: string
       publish-transaction-override:
         description: "Allow only the matching durable transaction to resume"
         required: false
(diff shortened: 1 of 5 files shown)


kungfu-origin (dismissed):
Reviewed the exact recovery caller diff and workflow authority projection. The new input requires an exact lowercase 40-character controller SHA, the Buildchain recovery train resolves to 66d360b1a91d48df468c930521782a1839934deb, and the consumer Gate controller remains read-only and separately bound from the immutable product subject. No candidate payload or product build path changes.

kungfu-origin (approved):
Re-reviewed the exact head a5c3706b91f882ca58ffd15d60d4bc0ab8de4af6 after publication-control-plane root refresh. The recovery controller is separately exact-bound from the immutable product subject; the generated workflow authority projection matches the updated source surface. No candidate payload, product build path, release identity, or target tree changes.

PR #2600, opened 2026-08-08
chore(maintainability): provision transition review authority

kungfu-origin:
## Summary

Provision one Ed25519 public review authority on the protected branch so independently signed, exact code-complexity baseline transitions can be verified.

This PR contains no baseline transition and no private key material.


dongkeren (approved):
Reviewed as the bounded protected-branch authority provisioning prerequisite; no baseline transition or private key material is included.

PR #2740, opened 2026-08-09
chore(maintainability): rotate transition review authority

dongkeren:
## Summary

- add a second independently scoped Ed25519 review authority for protected complexity baseline transitions
- keep the existing authority valid; no baseline, budget, waiver, or runtime behavior changes in this PR
- prepare a separately reviewed cleanup transition without weakening protected baseline integrity

## Validation

- full staged repository gate passed
- complexity budget and all 18 governance tests passed
- Buildchain KFD evidence was regenerated at the exact candidate head

## Security

Only the public key is committed. The private key is held outside Git, will sign only the subsequent exact cleanup transition, and will be destroyed after that transition is admitted.

### DIFF
--- a/.buildchain/kfd/kfd-3/collaboration-interface.prebuild.json
+++ b/.buildchain/kfd/kfd-3/collaboration-interface.prebuild.json
@@ -6,7 +6,7 @@
   "supportLevel": "release",
   "source": {
     "cwd": ".",
-    "sourceSha": "1d50286260f6bc9d5044a7c3b0a84db63291e941",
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Independent exact-head review: 09208904c09e0ded5f84448260c3643f3d682fa7 adds only a public Ed25519 transition-review authority and exact KFD projection refresh. It does not alter the active baseline, budgets, waivers, or runtime behavior. Approval is limited to enabling the separately signed cleanup transition.

PR #2826, opened 2026-08-10
fix(product): keep reference KFX outside assembly

dongkeren:
## Summary

- declare the maintained GitHub webhook KFX suite as reference-only product material
- prune a reference-only suite before recursive product KFX discovery and assembly
- prove ordinary product KFX still assemble while the excluded reference members stay outside the release

## Validation

- `node --test product/scripts/dist.test.mjs` (44/44)
- `node scripts/alpha-promotion-preflight.mjs fast-sentinel --kind auditable-demo --out /tmp/runtime-hub-auditable-demo-sentinel.json`
- `node --check product/scripts/dist.mjs`
- `node --check product/scripts/dist.test.mjs`
- `git diff --check`

This repairs the exact Build failure observed at `product/package.json must declare every product-assembled KFX dependency` without turning installed-only reference members into product dependencies.

## Evolution Map impact

Evolution impact: none

### DIFF
--- a/extensions/github-webhook-reference-suite/package.json
+++ b/extensions/github-webhook-reference-suite/package.json
@@ -3,5 +3,8 @@
   "version": "0.1.0",
   "description": "Maintained GitHub webhook ingress, evidence view, and optional Dogfood Finding bridge",
   "license": "Apache-2.0",
-  "private": true
+  "private": true,
+  "kungfuProduct": {
+    "assembly": "reference-only"
+  }
 }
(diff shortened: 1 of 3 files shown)


kungfu-origin (approved):
Independent review: the patch adds an explicit reference-only assembly boundary, prunes the complete suite subtree before product discovery, retains ordinary product KFX coverage, and includes a regression proving the reference members cannot enter release assembly. The exact three remote blobs match the locally tested files; DCO and protected checks remain authoritative.

PR #2858, opened 2026-08-10
fix(release): align transport rendition contract

dongkeren:
## Summary

- align the auditable-demo 720p rendition with the Buildchain v3 native contract (150x28)
- fail fast in Alpha preflight when the rendition vocabulary drifts
- add a regression test for the obsolete 100-column profile

## Root cause

Exact-dev self-hosted build run 31382604081 stopped before platform jobs because `.buildchain/auditable-demo-transport-smoke.json` still declared 720p columns=100, while the pinned Buildchain v3 validator requires 150 columns for both native profiles. The existing workflow test asserted the stale value and the fast sentinel did not validate rendition profiles.

## Validation

- targeted JS tests: 28/28 passed
- auditable-demo fast sentinel: passed
- `./shifu check:source`: passed
- staged pre-commit source gate: passed

No release, tag, or Alpha publication is requested.

### DIFF
--- a/scripts/alpha-promotion-preflight.mjs
+++ b/scripts/alpha-promotion-preflight.mjs
@@ -77,6 +77,24 @@ const NON_REUSABLE_EVIDENCE = [
   'signing',
 ];
 const EXACT_BUILDCHAIN_SHA = /^[0-9a-f]{40}$/u;
+const BUILDCHAIN_V3_NATIVE_RENDITIONS = [
+  {
+    id: '1080p',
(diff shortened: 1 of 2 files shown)


kungfu-origin (approved):
Independent review: the patch is limited to the stale Buildchain v3 rendition declaration plus an exact fail-fast sentinel and negative regression. The expected 150x28 720p profile matches the pinned validator; no publication or ADR-state behavior is added.

PR #2915, opened 2026-08-10
fix(ci): continue Linux-only auditable demo generation

dongkeren:
## Summary
- let the exact artifact resolver continue after intentional Windows sentinel skips
- keep the declarative demo job fail-closed on Linux build and resolver success
- pin workflow authority digests and add regression assertions

Supersedes #2913, whose historical PR merge graph remained dirty after a safe non-force update.

## Validation
- ./shifu test:auditable-demo
- ./shifu test:alpha-macos-overflow
- ./shifu kfd:buildchain:check
- ./shifu check:source
- ./shifu check:gate-catalog
- ./shifu check:release-publication-control-plane


kungfu-origin (approved):
Reviewed the exact four-file continuation repair and latest-dev reconciliation. The resolver and demo jobs remain fail-closed on build/resolver success, and targeted plus source gates are recorded.

PR #2943, opened 2026-08-11
fix(ci): isolate auditable demo generation

dongkeren:
## Outcome

Make auditable-demo generation a standalone, one-start Linux x86 workflow driven only by the declared scenarios, the exact product binary, Buildchain runtime, and the immutable renderer image.

## Boundary

- no Alpha Release Cut or KFD preflight
- no four-platform matrix or Windows sentinels
- no compiler cache, promotion, or release publication dependency
- manual and alpha/release triggers share the same generation implementation
- manual `gate-only` remains available for fast pipeline validation

## Validation

- `actionlint .github/workflows/auditable-demo.yml .github/workflows/build.yml`
- workflow-authority closed-world verification
- gate-catalog verification: 38 gates, 7 profiles, 28 structured invocations, 37 workflows
- 50 targeted workflow/cache/authority tests
- full staged pre-commit gate
- agent-120 one-start full generation at source `4930543f1fa171519181a7d9963c9ab230895911`: three demos, native 1920x1080 and 1280x720 captures, all fill ratios 1, all six capture roots distinct by resolution, RC=0

Exact contracts:

- renderer: `ghcr.io/kungfu-systems/build-images/demo-renderer@sha256:3a49708163fedaaabe07b45bba910026a1828151b5d4e9bbdaf0d62e75c927c1`
- Buildchain: `41d4dce2f38aa4821c794b49dcfb2bcf59d0984a`

PR #2946, opened 2026-08-11
fix(ci): isolate auditable demo generation

dongkeren:
## Outcome

Makes auditable-demo animation refresh a standalone, declarative, one-start Linux x86 workflow. It consumes the same-run product binary and an immutable Buildchain renderer image, then renders all registered scenarios at native 1080p and native 720p.

## Boundaries

- no Alpha Release Cut dependency
- no KFD qualification dependency
- no four-platform build dependency
- one Linux x64 product build only
- manual gate-only/full modes and optional media materialization share the same workflow
- workflow is explicitly registered as non-publication

## Verification

- exact one-commit linear branch based on dev
- full staged repository gate passed
- actionlint and workflow authority checks passed
- agent-120 one-start three-demo rendering passed
- all three demos produced separate native 1920x1080 and 1280x720 captures with fillRatio=1

Supersedes #2943, whose content was correct but whose inherited branch history could not be replayed by protected merge governance.


kungfu-origin (approved):
Reviewed as the linear replacement for #2943. The diff is confined to the standalone Linux x86 auditable-demo workflow, declarative animation surfaces, authority bindings, documentation, and closed-world contract updates. Proceed through protected checks and merge queue.

PR #2947, opened 2026-08-11
fix(ci): isolate auditable demo generation

dongkeren:
## Outcome

Makes auditable-demo animation refresh a standalone, declarative, one-start Linux x86 workflow. It consumes the same-run product binary and an immutable Buildchain renderer image, then renders all registered scenarios at native 1080p and native 720p.

## Boundaries

- no Alpha Release Cut dependency
- no KFD qualification dependency
- no four-platform build dependency
- one Linux x64 product build only
- manual gate-only/full modes and optional media materialization share the same workflow
- workflow is explicitly registered as non-publication

## Verification

- exact one-commit linear branch on latest dev 918ecabb79
- full staged repository gate passed at 13f8fe399dbbf01bab25613a0260941b83e1bd55
- actionlint and workflow authority checks passed
- agent-120 one-start three-demo rendering passed
- all three demos produced separate native 1920x1080 and 1280x720 captures with fillRatio=1

Supersedes #2946 and #2943; both preserve audit evidence for earlier moving-base/history attempts.


kungfu-origin (dismissed):
Reviewed as the final latest-base, fast-forward-only replacement. One signed-off commit, complete local staged gate, standalone Linux x86 animation boundary, and exact ADR-neutral declaration are present.

kungfu-origin (approved):
Re-reviewed exact head 26abc8e6724d6079993dd2b9df8df79e33999675. The only follow-up commit aligns the closed-world workflow-count test from 36 to 37; the targeted suite passes 24/24 and the full staged gate passes.

PR #2809, opened 2026-08-10
fix(product): qualify installer first-run paths

dongkeren:
## Summary

Qualify the standalone CLI and native provider first-run path across packaged products. This preserves an interactive terminal for first-run provider prompts, keeps Python bytecode out of the macOS application bundle, signs every V8 host with the required JIT profile, and aligns affected-native diagnostics with the pinned runtime.

## Related issue

None.

## Changes

- preserve inheritable standard I/O when the native Node worker launches provider-native UI
- add PTY and terminal-contract regressions for first-run interaction
- set `PYTHONPYCACHEPREFIX` before direct native worker startup and qualify the packaged cache boundary
- require the macOS JIT signing profile for every V8 host in standalone CLI and GUI layouts
- align affected-native runtime selection and regenerate source-bound qualification authority

## Verification

- `./shifu check:source`
- `./shifu check`
- pre-commit staged gate, including Rust clippy/workspace tests, KFD evidence checks, and documentation contracts
- exact-base merge precheck against `dev/v4/v4.0` at `80780651c94471ad135a231af9bfb20e087bd528`

## [Evolution Map](../docs/evolution/README.md) impact

Evolution impact: extends

## Governance risk check

Does this PR touch any of these boundaries?

- [ ] credentials, tokens, secrets, or private logs
- [x] provider APIs, CLIs, OpenTelemetry, billing, quota, or usage attribution
- [ ] official hosted or managed services
- [ ] official branding, package names, release identity, or domains
- [x] release evidence, provenance, package publishing, or deployment surfaces
- [ ] none of the above

The provider boundary is limited to terminal inheritance and first-run interaction. The release boundary is limited to packaging/signing/cache qualification and regenerated source-bound evidence; no package is published by this PR.

## Checklist

- [x] Commits are signed off (DCO)
- [x] Documentation updated if behavior changed

PR #3102, opened 2026-08-14
fix(delivery): pin non-family Warrant runtime

dongkeren:
## Summary

- pin the complete protected Warrant runtime suite to Buildchain `888e87fec88cad41d225c32615e80a87744d4b6d` from protected PR #2809
- update the queue authority contract, generated workflow authority projection, and exact-pin tests together
- reconcile stale current-base KFD evidence and its support-matrix hashes so the existing protected source gate remains internally coherent

## Why

The previously pinned controller rejected an otherwise fully qualified exact Warrant solely because a non-family candidate had no active family lease context. Buildchain #2809 removes only that unconditional rejection while retaining pending/readback enforcement whenever family context is supplied.

## Validation

- focused Warrant tests: 38/38
- `./shifu test:gate-latency`: 140/140
- workflow authority check and Kungfu Gate catalog: passed
- full staged commit gate: passed, including Buildchain KFD evidence, KFD support matrix, docs, invariants, and negative fixtures

Goal: `2026-08-12-adopter-category-profiles-and-universal-delivery-gate`

## Evolution Map impact

Evolution impact: none


kungfu-origin (dismissed):
Independent review: exact Buildchain pin is propagated across protected Warrant consumers, contract/tests/projection remain closed-world aligned, and the KFD evidence refresh is bounded to current-base source roots.

kungfu-origin (approved):

kungfu-origin (approved):

PR #3116, opened 2026-08-14
fix(work): admit verified compatible qualified core

dongkeren:
Merge fix/qualified-core-assignment-admission-compatibility-repair into dev/v4/v4.0 (kungfu-systems zone dual-account PR flow).

PR #3145, opened 2026-08-15
fix(work): admit verified compatible qualified core

dongkeren:
Merge fix/qualified-core-assignment-admission-compatibility-repair-linear-r2 into dev/v4/v4.0 (kungfu-systems zone dual-account PR flow).

Supersedes #3116: identical implementation cut, rebuilt as a protected-base-linear topology because GitHub merge queue reported merge_conflict for the prior non-linear head.


kungfu-origin (approved):
Independent review: exact head 17ad2e7caff239592703aec066ea57938f8e36f4 is a protected-base-linear two-commit recovery. The implementation matches the previously qualified cut; the KFD binding commit updates only exact source and evidence roots. FIT for protected source, Warrant, and merge-queue qualification.

PR #3148, opened 2026-08-15
fix(work): admit verified compatible qualified core

dongkeren:
Merge fix/qualified-core-assignment-admission-compatibility-repair-linear-r3 into dev/v4/v4.0 (kungfu-systems zone dual-account PR flow).

Supersedes #3145: identical implementation cut, freshly replayed onto protected base 32642e22df901a569651461e92cb6b467dd57c36 after exact delivery-contract evidence showed the r2 cut could not replay commit db96b6f910 onto the moved base. The new topology is strictly linear: protected base -> implementation b77966a0db -> exact-head KFD binding 78ed17662d.

PR #3155, opened 2026-08-15
fix(ci): preserve proof delta during attempt sealing

dongkeren:
## Summary

Preserve the admitted `dev-delta-plan` when sealing a reused affected-native proof on an advanced `dev/v4/v4.0` base.

This fixes the delivery-attempt failure observed while queueing PR #3148 in run 31871835473, where proof admission verified dependency attribution but the later seal step omitted that evidence and failed with `dependency-attribution-unknown`.

## Changes

- pass the admitted delta plan from the affected-native workflow into `seal-attempt`
- consume the optional delta plan when reconstructing the proof bundle
- cover the workflow/CLI contract in the gate-latency test suite
- refresh generated qualification and publication-surface roots

## Verification

- `./shifu test:gate-latency` — 142/142 passed
- `./shifu check:source` — Node 1130 passed / 11 skipped; Python 40 + 134 passed; desktop 73 passed
- commit hooks and generated-surface checks passed

## Architecture decision

No ADR required. This preserves an already verified proof-attribution input across two stages of the existing delivery workflow; it does not change an architecture contract.

## Evolution impact

None.

## Governance checklist

- [x] Release evidence and provenance paths checked
- [x] DCO sign-off present
- [x] Documentation/generator outputs refreshed where required


kungfu-origin (approved):
Independent review: verified the workflow preserves the already-admitted dev delta plan only for reused proofs, the seal-attempt CLI consumes the same optional input, generated authority roots were refreshed, and regression coverage binds both sides of the contract. Local full source acceptance and gate-latency validation are recorded on the PR.

PR #3156, opened 2026-08-15
fix(delivery): preserve reuse attribution through queue admission

dongkeren:
## Summary
- carry the already-admitted dev delta plan through both affected-native delivery-attempt sealing and cache-promotion authority sealing
- bind that delta plan into the cache authority so downstream verification preserves dependency attribution and remains fail-closed
- publish Warrant native proof under the distinct `Delivery Warrant native proof / linux` status context so it cannot shadow the app-owned required check
- refresh the closed-world workflow and publication authority roots

## Regression proof
- reproduced merge-group run 31894447782: delivery-attempt proof and Integration Delivery Proof succeeded, then cache-promotion authority rejected the same reused proof as `dependency-attribution-unknown`
- targeted affected-native tests passed (46 tests)
- code-complexity budget and gate-catalog checks passed
- full `./shifu check:source` passed

No required check is bypassed, and no publication is introduced.

## Evolution Map impact

Evolution impact: none


kungfu-origin (dismissed):
Reviewed the exact five-file change. The workflow carries only the already-admitted delta plan into the second verification; fail-closed behavior remains intact. Generated workflow/publication authority roots and regression coverage match the production change.

kungfu-origin (dismissed):
Reviewed exact head eedb985762ccfdbbe1682b864ab723700055cb49. The seven-file patch preserves admitted delta attribution through attempt/cache authority verification, keeps fail-closed proof checks, and separates the Warrant native status from the app-owned required check. Diff/check-source evidence is consistent.

PR #3161, opened 2026-08-15
fix(build): bind alpha workflow shell channel

dongkeren:
The exact merged source Build retry (run 31886634260) failed closed at the trust gate before the platform matrix because the reusable Buildchain workflow was pinned by an opaque commit SHA, so the shell could not prove its channel or major against runtime v3-alpha.

This repair binds the caller to the official v3-alpha workflow shell and updates the checked authority, bindings, rehearsal contract, publication control plane, and impacted tests to treat that auditable release-channel ref as the workflow-shell identity. It does not add an override or weaken the fail-closed gate.

Verification:
- 71 impacted tests pass
- workflow authority closed world verified
- gate catalog: 38 gates, 7 profiles, 28 structured invocations, 38 workflows aligned
- publication control plane qualifies
- commit hook full source validation passes
- native run gate returned ok:true

ADR delivery / release declaration:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -180,9 +180,9 @@ jobs:
   build:
     needs: [preflight, windows-fast-sentinel, auditable-demo-fast-sentinel]
     if: ${{ always() && needs.preflight.result == 'success' && needs.windows-fast-sentinel.result == 'success' && needs.auditable-demo-fast-sentinel.result == 'success' && !inputs.fast-sentinels-only && fromJSON(inputs.macos-overflow-request-json || '{}').mode != 'control' }}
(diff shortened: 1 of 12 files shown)


kungfu-origin (approved):
Independent review: the change is narrowly scoped to restoring the Buildchain channel/major proof boundary. The official v3-alpha shell now agrees with the existing v3-alpha runtime and consumer lock; authority registries, rehearsal contract, publication control plane, and negative tests move together. No bypass, fallback, publication behavior, or runtime override is introduced. Reviewed exact head 559709df80e1ffb981427ad74d6e9e089b0b8c37.

PR #3166, opened 2026-08-15
fix(release): reseal Alpha.2 attestation policy

dongkeren:
Bind the retained Alpha.2 candidate artifact identity to the detached Buildchain signer while keeping the GitHub provider source identity separate.

Validation:
- 26 promotion rehearsal tests passed
- workflow gate catalog: 38 gates, 7 profiles, 28 invocations, 38 workflows
- repository pre-commit gates passed

Candidate artifacts and release source are unchanged; this is release-tail control only.


kungfu-origin (approved):
Independent review: the patch changes only Alpha.2 release-tail workflow bindings and the generated authority projection; retained candidate source and artifact identity remain unchanged. Local rehearsal and gate-catalog evidence are coherent.

PR #3270, opened 2026-08-17
fix(release): bind Buildchain identity across merge groups

dongkeren:
Successor delivery for goal `2026-07-29-kungfu-terminal-review-remediation`.

Exact candidate:
- protected base: `0dc81a5dc5ebe043b7c7d67158bb33009088f653`
- head: `b395c01ab63794ad37266fa568de31fcc6c45e11`
- tree: `2804d0d37e9d65aa97fb364dca4a551692c8eece`
- scope: 23 files, 95 insertions, 51 deletions

The six-commit successor preserves immutable Buildchain caller/runtime identity across pull_request and merge_group execution, keeps portable documentation rooted by immutable source commit/tree, and restores the hidden Profile KFX fixture without exposing it as a public CLI surface.

Local qualification:
- Project Cut: `qualified`; fork `0dc81a5dc5ebe043b7c7d67158bb33009088f653`; replayed commits `6`; composition changed `false`; candidate tree `2804d0d37e9d65aa97fb364dca4a551692c8eece`
- `./shifu check:source`: passed
- Node tests: 1132 total, 1121 pass, 11 expected skips, 0 fail
- work-state: 40/40; runtime upgrade: 134/134; desktop: 73/73
- TypeScript and Biome: passed
- `./shifu kfd check --json`: `ok:true`, bounded-support
- zero-write tracked root: `sha256:6c52fc09b1da891745ba72f99800b12ed75e0347fb54a38a1986672a7cdb1948`
- zero-write untracked root empty: `sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`

This PR supersedes PR #3268 after protected base `0dc81a5dc5` made the old published history conflict during Project Cut replay. The successor was rebased locally and published under a new branch without force-pushing or reusing old review evidence. It requires fresh exact-head Core evidence, independent `kungfu-origin` review, non-preemptive Delivery Warrant admission, merge queue completion, post-merge exact-SHA qualification, and a fresh terminal-review attestation.

PR #3273, opened 2026-08-17
feat(release): cut temporal admission to Fact authority

dongkeren:
Merge `feature/fact-only-temporal-admission-r3` into `dev/v4/v4.0` through the kungfu-systems dual-account protected delivery flow.

This exact-base successor supersedes #3271 after protected base #3270 moved to `04c804c31b89aaf31d0763416cda15d29e5cc4f0`. The four substantive Project Cut commits replay cleanly at immutable source head `81fd211137925f7c69d2881b9c65698df862e573`; the prior intent-refresh commit is already present in the protected base.

This change makes verified temporal provenance and Buildchain compatibility Facts the sole normal admission authority. Digest projections remain deterministic but non-authoritative, while rollback is independently sealed, explicitly invoked, and ineligible for normal release decisions.


kungfu-origin (approved):
Independent exact-head review of 81fd211137925f7c69d2881b9c65698df862e573 against protected base 04c804c31b89aaf31d0763416cda15d29e5cc4f0: the four substantive Project Cut commits replay without composition drift; generated projections are canonical; temporal admission remains Fact-only and rollback remains offline and non-authoritative.

PR #3314, opened 2026-08-19
fix(runtime): recover ambiguous Assignment writes

dongkeren:
## Summary

- add an explicit public recovery plan and resolution path for interrupted Assignment Runtime writes whose authority outcome is unknown
- fence recovery by basis root, command root, revision, idempotency key, actor, reason, and evidence roots
- persist content-addressed recovery receipts and tombstone the original command so it can never be replayed
- keep authority outcome truthfully unknown and never invoke Work Control during local projection recovery

## Verification

- full ./shifu check:source passed on exact head 5a2d837785646a82b0532e1d5ad84791ad731f53
- Assignment Runtime JavaScript/GUI, Python, and runtime-service suites passed
- Python structure ratchet reports blocking=0
- Python type baseline passes for 248 source files

## ADR delivery / release declaration

Evolution impact: extends

Goal: 2026-08-18-kungfu-mainline-quality-convergence


kungfu-origin (approved):
Independent exact-head review complete: fail-closed unknown-outcome recovery preserves authority uncertainty, fences local resolution by exact provenance, prevents original-command replay, and is covered by schema plus restart/idempotency tests.

PR #3415, opened 2026-08-24
fix(release): preserve sealed KFD-10 evidence boundary

dongkeren:
## Summary

Preserve the sealed candidate's declared KFD-10 evidence boundary during Alpha recovery. Legacy candidates that predate the specialized runtime-warrant adopter witness no longer fail because a newer publication controller expects a post-cut source file, while candidates that declare the specialized witness continue to fail closed on any declaration drift.

## Verification

- `./shifu release:promotion:rehearse`
- `./shifu check:source`

## Evolution Map impact

Evolution impact: none

## Governance risk check

- [x] none of the above

## Checklist


kungfu-origin (dismissed):

kungfu-origin (approved):
Reviewed the exact de0cbe6081 head. The source-only exception is narrowly bound to deliveryClass=source-only plus a completed skipped provisional job; native-required delivery remains fail-closed. Workflow authority and publication roots were refreshed through the repository generators, and the full source gate passed locally.

PR #3607, opened 2026-08-31
fix(runtime): bound hidden-state retention

dongkeren:
## Summary

- fail closed runtime-image GC when implicit ownership cannot prove an active generation
- bound finalized provider logs by age and bytes while preserving unfinalized attempts
- lower default Codex log verbosity and document the maintenance contract

## Evolution Map impact

Evolution impact: none

## Validation

- `./shifu check:source`
- `node ./framework/core/.gyp/run-tests.mjs test:agent-console-contract`

## Governance risk check

- [x] none of the above

## Checklist

- [x] No secrets, private logs, generated build output, or local state included
- [x] Active, selected, rollback, and unknown-ownership state fails closed


kungfu-origin (approved):
Reviewed the bounded retention, fail-closed ownership handling, tests, and DCO identity.

PR #3611, opened 2026-08-31
fix(profile): relocate exact bundled lifecycle source

dongkeren:
## Summary

Keep Work Control Profile authority stable when an installed product image moves or is replaced.

## Related issue

None.

## Changes

- resolve an explicitly supplied bundled Profile source when it has the same Profile ID and exact content root as the lifecycle authority;
- recover a missing recorded Profile path only from the current bundled extension root, with the same exact ID and root checks;
- keep arbitrary extension-root search, root drift, and Profile substitution fail-closed;
- cover both rollback-retained and replaced-image layouts in the lifecycle and composition tests.

## Verification

- `./shifu check:source`
- 45 Work Authority tests
- 64 Profile SDK and composition tests
- 23 Work Control Profile tests
- preview distribution build `20260831T171228Z-d9ad86858`
- local preview promotion and installed-runtime exact-root verification

## Evolution Map impact

Evolution impact: none

## Governance risk check

- [x] none of the above

## Checklist

- [x] Documentation is not required because the external behavior and authority contract are unchanged

### DIFF
--- a/framework/core/src/python/kungfu/_profile_composition/support.py
+++ b/framework/core/src/python/kungfu/_profile_composition/support.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+import os
 from pathlib import Path
 from typing import Any, Mapping, NoReturn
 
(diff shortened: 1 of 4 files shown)


kungfu-origin (approved):
Independent exact-head review: the change preserves Profile ID and exact-root authority, narrows relocation to the explicit/current bundled source, keeps arbitrary discovery and root drift fail-closed, and includes positive plus negative regression coverage. Approved at d9ad86858235d591dda9591ad0644a6242c30241.

